PR #384, opened 2026-07-07
Keep dropdown body layout intact under global [popover] styles

acusti:
## Summary

Now that `.uktdropdown-body` renders in the top layer via the Popover API, it
matches a consuming app's global `[popover]` styles — and since
`.uktdropdown-body` and `[popover]` are both a single selector unit (`0,1,0`),
they tie on specificity and **source order decides**. We inject our `<style>`
via React precedence and can't guarantee we load last, so a consumer's
`[popover] { padding: … }` can win and break the dropdown's box geometry.

Surfaced in our own Storybook: `InputText.css` had a global `[popover]` rule
(`padding: 25px 30px`) that leaked onto every dropdown body (Dropdown's stories
also import that CSS).

## Fix

**Package — harden the layout, stay themeable.** Move the UA-popover box resets
(`inset` / `block-size` / `margin` / `border` / `padding`) out of
`.uktdropdown-body` into a `div.uktdropdown-body` rule. That's specificity
`(0,1,1)` — the *lightest* bump that still beats a bare `[popover]` `(0,1,0)`,
chosen deliberately over `.uktdropdown-body[popover]` `(0,2,0)` to keep the
override barrier as low as possible. Cosmetics (`background`, `box-shadow`,
`color`) stay at `(0,1,0)` so consumers can still theme the body.

**Docs — stop the bleed at the source.** The demo `[popover]` rule shouldn't be
global; scoped it to an `input-demo-popover` class on the two InputText demo
popovers (their appearance is unchanged).

## Verification (Storybook)

- The dropdown body's computed `padding` is now `0px` (was `25px 30px`).
- With a global `[popover] { padding: 25px 30px; border: 3px solid red }`
  injected at runtime, the body's `padding` and `border` **stay `0px`** —
  `div.uktdropdown-body` `(0,1,1)` wins over `[popover]` `(0,1,0)`.
- `tsc` / `lint` / `format` clean; `@acusti/dropdown` build passes; 70/70 tests.

Independent of the other open v1 PRs; branches off `main`. **No separate
changeset:** the top-layer popover rendering this hardens is still unreleased,
so the consumer-facing note is folded into the existing
`dropdown-top-layer-popover.md` (`major`) changeset rather than tracked as its
own entry.


.changeset/dropdown-harden-popover-body.md
acusti, line 1: i was thinking maybe we shouldn’t add a changeset for this and just let it be a part of the popover changes? or do you think it’s worth tracking separately?

acusti: Agreed — dropped the separate changeset and folded the one consumer-facing detail (that the resets hold against a consuming app's global `[popover]` styles) into the existing top-layer popover changeset, since the popover rendering is still unreleased. 13617436

acusti (commented):
LGTM

.changeset/dropdown-top-layer-popover.md
```diff
@@ -22,9 +22,11 @@ so consumer-side de-transforming workarounds are no longer needed.
...
 - Submenus still anchor to their parent item and escape the body’s
   `overflow: hidden` in the top layer; the body’s `z-index` is dropped
-  since the top layer handles stacking, and the UA popover styles
-  (`inset`/`margin`/`border`/`padding`/`color`) are reset so they can’t
-  override the anchor-positioning layout
+  since the top layer handles stacking, and the UA popover box resets
+  (`inset`/`block-size`/`margin`/`border`/`padding`) are applied at a
+  specificity that also overrides a consuming app’s global `[popover]`
+  styles, so neither the UA defaults nor a consumer’s popover CSS can break
+  the anchor-positioning layout
```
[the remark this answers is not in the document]
acusti: Reconciled by fixing the description, not the bump. The standalone `patch` changeset was removed and its note folded into the existing `dropdown-top-layer-popover.md`, which is `major` — correct, since the top-layer rendering this hardens is unreleased and ships as one v1 entry. Updated the PR description to drop the stale "patch changeset" line.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #394, opened 2026-07-15
Dropdown: flexible value/label — derive from children and { label, value } pairs

acusti:
## Summary

Decouples a dropdown item's stored value from its displayed label — for controlled, searchable dropdowns where the two differ (e.g. an id stored under a human-readable label). Two layered, additive, backward-compatible ways to supply the label:

1. **Derive from children** — when you render the items yourself, a bare string `value` shows the label of the child whose `data-ukt-value` matches it. No extra prop; the everyday default.
2. **`{ label, value }` value pair** — state the label explicitly (the same shape `onSubmitItem` reports back, so a controlled consumer can feed back what it received); the override when the derived text isn't right.

For a bare `value`, the label resolves as: explicit pair label → matching child's text → the raw identifier.

Previously `value` was a single string doing double duty: the searchable input's displayed text **and** the identity compared against each item's `data-ukt-value` for change detection. When value ≠ label those conflict — displaying the label forced `value` to _be_ the label, so `onSubmitItem` reported the label and consumers had to map it back to an id.

> An `options` prop (a data-driven body) was originally part of this PR and is split out into #397 to be weighed on its own.

## Details

- `value` normalizes to `valueIdentity` (drives change detection / no-op matching + `valueRef`) and `valueLabel` (the searchable input's `defaultValue`).
- The derived-from-children label is a whitespace-collapsed concatenation of the matching item's text leaves, approximating the rendered `innerText` — which is what the input already shows on select, so the closed-state and post-select labels come from the same place. A `{ label, value }` pair overrides it when that text isn't right.
- Fully backward compatible: existing string-`value` callers are unchanged.

## Testing

- 6 new tests: pair display, string backward-compat, identity-based no-op, children derivation, nested-text concatenation, and the pair override. Full suite: **244 passing**.
- Two `minor` changesets.
- Two Storybook stories: `LabelDerivedFromChildren` (zero-config) and `LabelDifferentFromValue` (the pair as an explicit override).
- `build` / `test` / `tsc` / `lint` / `format` all green.

### DIFF
--- a/packages/docs/stories/Dropdown.stories.tsx
+++ b/packages/docs/stories/Dropdown.stories.tsx
@@ -300,11 +300,41 @@ const COUNTRIES = [
     { label: 'United States', value: 'US' },
 ] as const;
 
+export const LabelDerivedFromChildren: Story = {
+    parameters: {
+        docs: {
+            description: {
+                story: 'When you render the items yourself, a bare string `value` displays the label of the child whose `data-ukt-value` matches it — so items whose value and label differ (an ISO country code shown as the country name) need no `{ value, label }` pair. Type to filter by the visible label; `onSubmitItem` reports { value, label }. The derived label is the matching item’s text — pass a `{ value, label }` pair to override it when that text isn’t what you want shown in the input.',
+            },
+        },
+    },
+    render() {
+        const [country, setCountry] = React.useState('US');
+        return (
+            <Dropdown
+                isSearchable
+                label="Country"
+                onSubmitItem={({ value }) => setCountry(value)}
+                placeholder="Search countries…"
+                value={country}
+            >
+                <ul>
+                    {COUNTRIES.map(({ label, value }) => (
+                        <li data-ukt-value={value} key={value}>
+                            {label}
+                        </li>
+                    ))}
+                </ul>
+            </Dropdown>
+        );
+    },
+};
+
 export const LabelDifferentFromValue: Story = {
     parameters: {
         docs: {
             description: {
-                story: 'When an item’s stored value differs from its display label — here an ISO country code stored under the country name — pass `value` as a `{ value, label }` pair. The `value` drives change detection and matching against each item’s `data-ukt-value`, while the `label` is shown in the searchable input (and is what you type to filter). `onSubmitItem` reports the same `{ value, label }` shape back, so a controlled consumer feeds it straight into state — no mapping a label back to an id. The selected code is stored and echoed below; a bare string `value` still works when an item’s value and label are the same.',
+                story: 'A `{ value, label }` pair states the displayed label explicitly, overriding the text that would otherwise be derived from the matching item. Here the list items carry the ISO code for scannability, but the collapsed input shows just the country name. `value` drives change detection and matching against each item’s `data-ukt-value`, and `onSubmitItem` reports the same `{ value, label }` shape back, so a controlled consumer feeds it straight into state.',
             },
         },
     },
@@ -325,7 +355,7 @@ export const LabelDifferentFromValue: Story = {
                     <ul>
                         {COUNTRIES.map(({ label, value }) => (
                             <li data-ukt-value={value} key={value}>
-                                {label}
+                                {label} · {value}
                             </li>
                         ))}
                     </ul>
--- a/packages/docs/stories/Dropdown.stories.tsx
+++ b/packages/docs/stories/Dropdown.stories.tsx
@@ -290,6 +290,54 @@ export const SearchableAndAllowCreate: Story = {
     },
 };
 
+const COUNTRIES = [
+    { label: 'Australia', value: 'AU' },
+    { label: 'Brazil', value: 'BR' },
+    { label: 'Canada', value: 'CA' },
+    { label: 'Germany', value: 'DE' },
+    { label: 'Japan', value: 'JP' },
+    { label: 'United Kingdom', value: 'GB' },
+    { label: 'United States', value: 'US' },
+] as const;
+
+export const LabelDifferentFromValue: Story = {
+    parameters: {
+        docs: {
+            description: {
+                story: 'When an item’s stored value differs from its display label — here an ISO country code stored under the country name — pass `value` as a `{ value, label }` pair. The `value` drives change detection and matching against each item’s `data-ukt-value`, while the `label` is shown in the searchable input (and is what you type to filter). `onSubmitItem` reports the same `{ value, label }` shape back, so a controlled consumer feeds it straight into state — no mapping a label back to an id. The selected code is stored and echoed below; a bare string `value` still works when an item’s value and label are the same.',
+            },
+        },
+    },
+    render() {
+        const [country, setCountry] = React.useState('US');
+        const selected = COUNTRIES.find(({ value }) => value === country);
+        return (
+            <div style={{ display: 'flex', flexDirection: 'column', gap: '0.75rem' }}>
+                <Dropdown
+                    isSearchable
+                    label="Country"
+                    onSubmitItem={({ value }) => setCountry(value)}
+                    placeholder="Search countries…"
+                    value={
+                        selected ? { label: selected.label, value: selected.value } : ''
+                    }
+                >
+                    <ul>
+                        {COUNTRIES.map(({ label, value }) => (
+                            <li data-ukt-value={value} key={value}>
+                                {label}
+                            </li>
+                        ))}
+                    </ul>
+                </Dropdown>
+                <p style={{ margin: 0 }}>
+                    Stored value: <code>{country}</code>
+                </p>
+            </div>
+        );
+    },
+};
+
 export const CSSValueInputTrigger: Story = {
     args: {
         allowCreate: true,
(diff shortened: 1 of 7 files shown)


packages/dropdown/src/Dropdown.test.tsx
```diff
@@ -730,6 +730,185 @@ describe('@acusti/dropdown', () => {
...
+                <Dropdown
+                    onSubmitItem={handleSubmitItem}
+                    value={{ label: 'Warm & Welcoming', value: 'warm' }}
+                >
+                    Voice
+                    <ul>
+                        <li data-testid="warm" data-ukt-item data-ukt-value="warm">
+                            Warm & Welcoming
+                        </li>
+                        <li data-testid="bold" data-ukt-item data-ukt-value="bold">
```
acusti: this is redundant, any DOM element with a `data-ukt-value` is treated as a `data-ukt-item`. is there a reason to encode that redundancy in this test?

packages/docs/stories/Dropdown.stories.tsx
```diff
@@ -290,6 +290,107 @@ export const SearchableAndAllowCreate: Story = {
...
+            </Dropdown>
+        );
+    },
+};
+
+export const LabelDifferentFromValue: Story = {
+    parameters: {
+        docs: {
+            description: {
+                story: 'A `{ value, label }` pair states the displayed label explicitly, overriding the text that would otherwise be derived from the matching item. Here the list items carry the ISO code for scannability, but the collapsed input shows just the country name. `value` drives change detection and matching against each item’s `data-ukt-value`, and `onSubmitItem` reports the same `{ value, label }` shape back, so a controlled consumer feeds it straight into state.',
```
acusti: dumb nit, but for whatever reason, for here and elsewhere, i prefer `{ label, value }`

packages/dropdown/src/Dropdown.test.tsx
```diff
@@ -730,6 +730,185 @@ describe('@acusti/dropdown', () => {
...
+            expect((screen.getByRole('textbox') as HTMLInputElement).value).toBe(
+                'Font Weight - 400',
+            );
+        });
+
+        it('lets a { value, label } pair override the label derived from children', () => {
+            render(
+                <Dropdown isSearchable value={{ label: 'Toasty', value: 'warm' }}>
+                    <ul>
+                        <li data-ukt-value="warm">Warm &amp; Welcoming</li>
```
acusti: the inconsistent `&amp;` here vs the plain `&` above (e.g. line 741) is confusing

packages/dropdown/src/Dropdown.tsx
```diff
@@ -177,6 +205,46 @@ export default function Dropdown(props: Props) {
...
+            visit(childProps.children);
+        });
+    };
+    visit(children);
+    return label;
+}
+
+// Concatenate the text (string/number leaves) of a React node, approximating
+// the rendered innerText the dropdown uses as an item’s label.
+function getNodeText(node: ReactNode): string {
```
acusti: this should go in `./helpers.ts`

packages/dropdown/src/Dropdown.tsx
```diff
@@ -177,6 +205,46 @@ export default function Dropdown(props: Props) {
     return <RootDropdown {...props} />;
 }
 
+// Find the item in `children` whose data-ukt-value matches `value` and return
+// its text as the label, so a bare identifier value shows the matching item’s
+// label without a separate options/label lookup. Returns undefined if no item
+// matches (the caller then falls back to the identifier itself). A
+// { value, label } value overrides this when the derived text isn’t right.
+function getLabelFromChildren(children: ReactNode, value: string): string | undefined {
```
acusti: this should also go in `./helpers.ts`

packages/dropdown/src/Dropdown.test.tsx
acusti: Fixed in `a59b8d36` — dropped the redundant `data-ukt-item` from this PR's test items (left the pre-existing tests that use the same pattern alone). Applied the same cleanup to the reveal tests on #396 (`19b5e940`).

packages/docs/stories/Dropdown.stories.tsx
acusti: Fixed in `744d05ba` — swept every `{ value, label }` in prose (TSDoc, README, changesets, story descriptions, and test names) to `{ label, value }`, matching the alphabetized shape the code already writes. PR title/body updated to match.

packages/dropdown/src/Dropdown.test.tsx
acusti: Fixed in `a59b8d36` — plain `&` in JSX text now, consistent with the neighboring tests.

packages/dropdown/src/Dropdown.tsx
acusti: Fixed in `d00f0bbd` — moved into helpers.ts alongside the other item/label helpers (arrow-const style to match the file; `getNodeText` stays module-local since only `getLabelFromChildren` is consumed outside).

packages/dropdown/src/Dropdown.tsx
acusti: Fixed in `d00f0bbd`, together with `getNodeText`.

packages/dropdown/src/helpers.ts
```diff
@@ -42,6 +42,49 @@ export const getItemLabel = (item: HTMLElement): string => {
...
+    let label: string | undefined;
+    const visit = (node: ReactNode) => {
+        if (label != null) return;
+        Children.forEach(node, (child) => {
+            if (label != null || !isValidElement(child)) return;
+            const childProps = child.props as {
+                children?: ReactNode;
+                'data-ukt-value'?: unknown;
+            };
+            if (childProps['data-ukt-value'] === value) {
```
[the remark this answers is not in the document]
acusti: Fixed in `c67eb05` — `getLabelFromChildren` now matches on `String(uktValue) === value` (guarded by `uktValue != null`), so a numeric `data-ukt-value={400}` derives its label against `value="400"`, consistent with the string-coerced `dataset.uktValue` used for selection. Derivation no longer falls back to the raw identifier for numeric values.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -215,6 +224,20 @@ function RootDropdown({
...
+    // item’s data-ukt-value; valueLabel is what a searchable dropdown shows in
+    // its input. For a bare identifier, the label is resolved from the matching
+    // child (data-ukt-value in the body), so children with distinct values and
+    // labels need no separate label; a { label, value } pair states it
+    // explicitly.
+    const valueIdentity = typeof value === 'string' ? value : value?.value;
+    const valueLabel =
+        typeof value !== 'string'
+            ? value?.label
+            : (getLabelFromChildren(children, value) ?? value);
```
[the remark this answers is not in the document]
acusti: Fixed in `94c81eb` — `valueLabel` is now computed only when `isSearchable`; otherwise it's `undefined`. Since it's read only as the searchable input's `defaultValue` (the `<input>` trigger), non-searchable dropdowns skip the `getLabelFromChildren` traversal entirely.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -55,11 +56,17 @@ export type Item = {
...
 };
 
-export type ItemPathEntry = { label: string; value: string };
+/**
+ * A { label, value } pair naming an item: `value` is the stored value (the
+ * item’s data-ukt-value) and `label` is its displayed text. Accepted by the
+ * value prop when an item’s value and label differ; also the shape of an
+ * Item’s path entries.
+ */
+export type ItemValue = { label: string; value: string };
```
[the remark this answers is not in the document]
acusti: ItemPathEntry was added in an alpha and removed a few alphas later, so this is completely semver-compatible and, in practical terms, safe churn. maintaining an alias for the old type export would be confusing and a mistake.

packages/dropdown/src/Dropdown.tsx
```diff
@@ -215,6 +224,22 @@ function RootDropdown({
...
+    // child (data-ukt-value in the body), so children with distinct values and
+    // labels need no separate label; a { label, value } pair states it
+    // explicitly. Only a searchable dropdown displays the label, so skip the
+    // per-render children walk entirely for anything else.
+    const valueIdentity = typeof value === 'string' ? value : value?.value;
+    const valueLabel = !isSearchable
+        ? undefined
+        : typeof value !== 'string'
+          ? value?.label
+          : (getLabelFromChildren(children, value) ?? value);
```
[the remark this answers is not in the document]
acusti: Fixed in `7688bce` — the body node (the second of two children, or the only child) is now bound once and used for both rendering and label derivation, so derivation walks exactly what `.uktdropdown-body` renders and a `data-ukt-value` inside a custom trigger can never match or shadow a body item.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -215,6 +230,22 @@ function RootDropdown({
...
```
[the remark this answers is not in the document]
acusti: Duplicate of the earlier thread on this line — fixed in `7688bce`: derivation now receives the body node (shared with rendering) instead of the full children tuple.

---

PR #403, opened 2026-07-21
Budget dropdown fill fallbacks to the position-try evaluation limit

### LINKED ISSUE #399: Dropdown body overflows off-screen at viewport-edge triggers when taller than the roomy side (position-try fill fallbacks not reached)
> [!NOTE]
> **Resolved in #403.** The root-cause diagnosis below (a `position-try-fallbacks` evaluation quirk where "only the first fill is evaluated," possibly an upstream Chromium bug) was **superseded** during the fix: the real cause is the CSS anchor positioning spec's implementation-defined limit on the position-options list length (floor of five, base placement included), not a Chromium bug. See the [resolution comment](https://github.com/acusti/uikit/issues/399#issuecomment-5040397979) for the corrected explanation and the shipped fix. Original report preserved below for history.

---

## Summary

When a `Dropdown` body is nearly as tall as the viewport and its trigger sits near a block-axis viewport edge, the body can render **off-screen** instead of flipping/shrinking to the roomy side. The `--uktdd-fill` `@position-try` fallbacks that are meant to rescue this case aren't reliably reached in Chromium — once the body exceeds the space on the edge side, the engine reverts to the primary placement and overflows the viewport.

Consumers currently have to work around it by capping `--uktdd-body-max-height` to a viewport fraction (e.g. `40dvh`). This issue documents the root cause and why the obvious in-library fixes don't work, so we can decide on a real mitigation.

## Environment

- `@acusti/dropdown@1.0.0-alpha.5`
- Chromium `141.0.7390.37` (headless); reproduces in desktop Chrome too
- Repro requires the body's natural height to be ≳ ~90% of the viewport height **and** the trigger within roughly `min-block-size` of a block-axis viewport edge (i.e. short viewports / very long menus).

## Repro

A menu with enough items to be ~viewport-tall, trigger pinned near the bottom edge, short viewport (e.g. `1200×420`). Consumer CSS with **no** height cap:

```css
.card-menu .uktdropdown-body {
  --uktdd-body-position-area: block-end span-inline-start;
  --uktdd-body-position-try-fallbacks:
    --uktdd-top-end, --uktdd-bottom-start, --uktdd-top-start;
  /* no --uktdd-body-max-height override */
}
```

- **Expected:** body flips above the trigger (there is ~384px of room above) and scrolls.
- **Actual:** body opens *below* the trigger at its clamped max height (~410px) and runs off the bottom of the viewport (`getBoundingClientRect().bottom ≈ 822` in a 420px-tall viewport).

## Diagnosis

This is a `position-try-fallbacks` **evaluation** quirk, not a math error in the shipped CSS. Isolating the fallbacks at the failing bottom-edge trigger (420px viewport):

- With **only** `--uktdd-top-end` as the fallback → the body shrinks to the 384px above and fits (`height: 384px`, on-screen). So the body *can* fit above.
- With the **full shipped list** (`…, --uktdd-fill, --uktdd-fill flip-inline, --uktdd-fill flip-block, --uktdd-fill flip-block flip-inline`) → the engine reverts to the primary placement and overflows.
- Bisecting the list: once the body exceeds the edge-side space, **only the *first* fill fallback is effectively evaluated**. The later fills — including the opposite-block `--uktdd-fill flip-block` that would fit — are never reached, so the engine falls back to the original (primary) position and overflows.

### In-library fixes that were tried and don't work

- **Reorder the appended fills** so opposite-block (`flip-block`) fills come first → fixes the bottom edge but **breaks the top edge** and flips every menu's preferred direction upward. It just relocates the failure to the other edge (the "only the first fill is tried" behavior is symmetric).
- **`position-try-order: most-height`** → no effect; the same faulty overflow evaluation defeats it (still reverts to primary, still off-screen).
- **Cap `max-block-size` to the roomy side dynamically**, e.g. `max-block-size: max(anchor(top), calc(100dvh - anchor(bottom)))` → `anchor()` is **not valid in sizing properties**, so the declaration is dropped and you get the default back. (`anchor-size()` is valid there but returns the anchor's own size, not the gap to the viewport edge.)

### What does work

Bounding the body height so the **named** fallbacks (which *do* fit at natural size) can land it on the roomy side:

- `--uktdd-body-max-height: ≤ 50dvh` is **provably safe** — the two block-axis sides of the anchor sum to ~100dvh, so the roomy side is always ≥ ~50dvh, and a body capped below that always fits after a flip.
- More generous caps (`60/75/85dvh`, `calc(100dvh − 48px)`) clear all *typical* trigger positions in testing but are **not** provably robust: a trigger positioned ~`min-block-size` (~30px) off the edge is exactly the zone where the fill is needed and fails, so it can still overflow there.

Sweep of a card-menu trigger from the top → bottom of the viewport (`▼` opens below, `▲` flips above, ❌ off-screen):

```
shipped (no cap), 420px viewport:   ▼ ▼ ▼ ▼ ▼   ❌ bottom edge + corners
--uktdd-body-max-height: 40dvh:     ▼ ▼ ▼ ▲ ▲   ✓ all on-screen
```

## Possible mitigations

1. **Lower the default `--uktdd-body-max-height`** (currently `calc(100dvh - var(--uktdd-body-buffer))`, buffer `10px`). A viewport-fraction default (≤ ~50dvh) is fully robust but caps *all* menus to ~half the viewport and makes them scroll sooner. Bumping `--uktdd-body-buffer` to ~40–48px clears the realistic cases without restricting normal menus, but isn't a total guarantee for every trigger geometry.
2. **Document `--uktdd-body-max-height` as the intended knob** for surfaces whose menu can be near-viewport-height at an edge, and leave the default generous. (This is what a downstream consumer landed on: a `40dvh` cap on the affected menus.)
3. If this is an upstream Chromium bug (later `position-try` fallbacks not evaluated once the box exceeds the edge-side space), track it — once fixed, the `--uktdd-fill` fallbacks would do their job and no cap would be needed. I wasn't able to test other engine versions to confirm whether it's version-specific.

acusti:
## Summary

Fixes #399 — a `Dropdown` body nearly as tall as the viewport could open off-screen at a block-axis viewport edge instead of flipping/scrolling, and consumers had to cap `--uktdd-body-max-height` to work around it.

**Root cause (empirically confirmed, and different from the issue's diagnosis):** it isn't a fallback-evaluation quirk in the fills — the CSS anchor positioning spec permits an implementation-defined limit on the length of position-options lists ("must be at least five"), and **Chromium evaluates exactly five options and silently ignores the rest**. The shipped list was 3 author fallbacks + 4 `--uktdd-fill` tactic variants = 7 options, so `--uktdd-fill flip-block` (6th) and `--uktdd-fill flip-block flip-inline` (7th) were unreachable dead code. That also explains why reordering the fills "fixed one edge but broke the other" (it moved a different option past the cliff) and why `position-try-order: most-height` had no effect.

## The fix

Replace the four appended tactic fills with two named options — `--uktdd-fill-bottom` and `--uktdd-fill-top` — bringing the default list to exactly five. Each fill pairs:

- a single-side `position-area` (`block-end` / `block-start`), which spans all inline tiles so the option can never be rejected for horizontal overflow, with
- an explicit `justify-self: anchor-center`, which centers the body over the trigger and shifts it inward as needed to stay on-screen. The explicitness is load-bearing: `span-all`'s *default* alignment centers the same way but skips the shift, so near a viewport corner the option would overflow and be rejected.

A fill can then only be rejected when its side is shorter than `--uktdd-body-min-height`, and the two block sides sum to the viewport, so at least one of the pair always fits — for every trigger position, including corners, with no height cap needed. The pair is appended via a new `--uktdd-body-fill-fallbacks` custom property so upward-opening dropdowns can flip its order to keep the last-resort fill on their preferred side.

Submenus get the same budget: author fallback, `--uktdd-fill`, `--uktdd-fill flip-inline`, then the two block-side fills (also exactly five).

**Also fixes a pre-existing submenu bug** surfaced while testing: the submenu rule's `--uktdd-body-gap: 0` override was unitless, making the fills' `calc(100% - var(--uktdd-body-gap) * 2)` block-size invalid (`<percentage> - <number>`), so every submenu fill option was evaluated at natural size and rejected — submenu fill fallbacks have never applied, and a too-tall submenu overflowed the viewport. Now `0px`.

## Behavior trade-off

In the fill placements the body spans the trigger (anchor-centered, shifted to fit) instead of keeping the primary placement's edge alignment. Fills only apply when the body fits nowhere at its natural size, so normal placement is unchanged. Consumer fallback lists must stay within the budget — at most 3 options in `--uktdd-body-position-try-fallbacks`, 1 in `--uktdd-submenu-position-try-fallbacks` — now documented in the README.

## Validation

Verified in headless Chromium 141.0.7390.37 (the version from the issue report) with a Playwright anchor-positioning harness using the real `Dropdown.css`:

- 90 body placements swept across bottom/top edges and left/right corners in a 1200×420 viewport, for the issue's end-aligned config, the default config, and a top-opening recipe (with flipped fill pair), with and without `--uktdd-body-gap` — all land fully on-screen; the issue's failing case now flips above and fills the roomy side
- short bodies confirm named fallbacks still govern natural-size placement
- submenu smoke tests: sideways placement and inline flip preserved; tall submenus now fill and scroll; near-bottom and corner cases stay on-screen
- the five-option limit itself confirmed with a micro-test (N failing dummy options before a fitting one: index 5 reached, index 6 never)

`bun run build`, dropdown tests (102 passed), and `bun run format:check` all pass.

https://claude.ai/code/session_019rR4n1tgs2TWm9CtSdGaqX

---
_Generated by [Claude Code](https://claude.ai/code/session_019rR4n1tgs2TWm9CtSdGaqX)_

### DIFF
--- a/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
+++ b/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
@@ -22,11 +22,14 @@ exactly five. Each fill pairs a single-side `position-area`
 rejected for horizontal overflow) with `justify-self: anchor-center`, which
 centers the body over the trigger and shifts it inward as needed to stay
 on-screen (the explicitness matters: `span-all`’s default alignment centers
-without that shift, so near a corner it would overflow and be rejected).
-Because a fill can only be rejected when its side is shorter than
-`--uktdd-body-min-height` and the two sides sum to the viewport, at least
-one of the pair always fits — for every trigger position, including
-corners, with no height cap needed. The pair is appended via the new
+without that shift, so near a corner it would overflow and be rejected). A
+fill is only rejected when its side is shorter than the body’s
+min-block-size floor — what sends a cramped side’s fill to the roomier
+opposite side — and each fill caps that floor at the worst-case larger side
+(half the viewport minus the trigger, the most a block-centered trigger can
+guarantee), so at least one of the pair always fits — for every trigger
+position, including corners and a `--uktdd-body-min-height` above half the
+viewport, with no height cap needed. The pair is appended via the new
 `--uktdd-body-fill-fallbacks` custom property, so upward-opening dropdowns
 can flip its order to keep the last-resort fill on their preferred side.
 
--- a/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
+++ b/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
@@ -6,18 +6,23 @@ Keep the last-resort fill placements within the `position-try` evaluation
 limit so a too-tall body flips and scrolls instead of opening off-screen
 (fixes #399)
 
-The CSS anchor positioning spec lets engines cap how many
-`position-try-fallbacks` options they evaluate, as low as five — and
-Chromium evaluates exactly five, silently ignoring the rest. The body’s
-previous list was three author fallbacks plus four `--uktdd-fill` tactic
-variants (seven options), so the block-flipped fills that rescue a
-near-viewport-height body at a block-axis viewport edge were never
-evaluated: the body opened toward the edge and ran off-screen, and
-consumers had to cap `--uktdd-body-max-height` to work around it.
+The CSS anchor positioning spec lets engines cap the position options list
+at an implementation-defined length with a floor of five — and the list
+includes the element’s base position, so only four fallbacks past it are
+guaranteed. Chromium evaluates five fallbacks past the base, silently
+ignoring the rest. The body’s previous list was three author fallbacks plus
+four `--uktdd-fill` tactic variants (seven fallbacks), so the block-flipped
+fills that rescue a near-viewport-height body at a block-axis viewport edge
+were never evaluated: the body opened toward the edge and ran off-screen,
+and consumers had to cap `--uktdd-body-max-height` to work around it.
 
 The four appended tactic fills are replaced by two new named options —
 `--uktdd-fill-bottom` and `--uktdd-fill-top` — bringing the default list to
-exactly five. Each fill pairs a single-side `position-area`
+exactly five fallbacks, Chromium’s evaluated budget. That is one past the
+spec’s guaranteed floor: an engine at the exact floor would evaluate only
+four and drop the second fill. The design is validated in Chromium — the
+engine whose limit bites — and untested at the floor, since no known engine
+sits there. Each fill pairs a single-side `position-area`
 (`block-end`/`block-start`, spanning all inline tiles so it can never be
 rejected for horizontal overflow) with `justify-self: anchor-center`, which
 centers the body over the trigger and shifts it inward as needed to stay
@@ -31,15 +36,20 @@ guarantee), so at least one of the pair always fits — for every trigger
 position, including corners and a `--uktdd-body-min-height` above half the
 viewport, with no height cap needed. The pair is appended via the new
 `--uktdd-body-fill-fallbacks` custom property, so upward-opening dropdowns
-can flip its order to keep the last-resort fill on their preferred side.
+can flip its order to keep the last-resort fill on their preferred side;
+submenus reuse the pair as their own final rescue, so the flip carries into
+a body’s submenus as well.
 
 In the fill placements the body now spans the trigger (anchor-centered,
 shifted to fit) instead of keeping the primary placement’s edge alignment.
 Consumer fallback lists must stay within the same budget: at most three
-options in `--uktdd-body-position-try-fallbacks`, and a single option in
+options in `--uktdd-body-position-try-fallbacks` (two, to stay within the
+spec floor), and a single option in
 `--uktdd-submenu-position-try-fallbacks` (the submenu list is now its
 author fallback, `--uktdd-fill`, `--uktdd-fill flip-inline`, then the two
-block-side fills — also exactly five).
+block-side fills — also exactly five fallbacks; unlike the body’s list it
+can’t shrink below five without losing side coverage, so submenu rescue
+inherently relies on engines evaluating past the spec floor).
 
 Also fixes the submenu rule’s `--uktdd-body-gap: 0` override: the unitless
 zero made the fill options’ `calc(100% - var(--uktdd-body-gap) * 2)`
--- a/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
+++ b/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
@@ -9,12 +9,14 @@ limit so a too-tall body flips and scrolls instead of opening off-screen
 The CSS anchor positioning spec lets engines cap the position options list
 at an implementation-defined length with a floor of five — and the list
 includes the element’s base position, so only four fallbacks past it are
-guaranteed. Chromium evaluates five fallbacks past the base, silently
-ignoring the rest. The body’s previous list was three author fallbacks plus
-four `--uktdd-fill` tactic variants (seven fallbacks), so the block-flipped
-fills that rescue a near-viewport-height body at a block-axis viewport edge
-were never evaluated: the body opened toward the edge and ran off-screen,
-and consumers had to cap `--uktdd-body-max-height` to work around it.
+guaranteed. In practice Chromium is the tightest, evaluating six options
+(five fallbacks past the base) and silently ignoring the rest, while Safari
+and Firefox evaluate many more. The body’s previous list was three author
+fallbacks plus four `--uktdd-fill` tactic variants (seven fallbacks), so
+the block-flipped fills that rescue a near-viewport-height body at a
+block-axis viewport edge were never evaluated: the body opened toward the
+edge and ran off-screen, and consumers had to cap `--uktdd-body-max-height`
+to work around it.
 
 The four appended tactic fills are replaced by two new named options —
 `--uktdd-fill-bottom` and `--uktdd-fill-top` — and the default author
--- a/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
+++ b/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
@@ -1,0 +1,47 @@
+---
+'@acusti/dropdown': minor
+---
+
+Keep the last-resort fill placements within the `position-try` evaluation
+limit so a too-tall body flips and scrolls instead of opening off-screen
+(fixes #399)
+
+The CSS anchor positioning spec lets engines cap how many
+`position-try-fallbacks` options they evaluate, as low as five — and
+Chromium evaluates exactly five, silently ignoring the rest. The body’s
+previous list was three author fallbacks plus four `--uktdd-fill` tactic
+variants (seven options), so the block-flipped fills that rescue a
+near-viewport-height body at a block-axis viewport edge were never
+evaluated: the body opened toward the edge and ran off-screen, and
+consumers had to cap `--uktdd-body-max-height` to work around it.
+
+The four appended tactic fills are replaced by two new named options —
+`--uktdd-fill-bottom` and `--uktdd-fill-top` — bringing the default list to
+exactly five. Each fill pairs a single-side `position-area`
+(`block-end`/`block-start`, spanning all inline tiles so it can never be
+rejected for horizontal overflow) with `justify-self: anchor-center`, which
+centers the body over the trigger and shifts it inward as needed to stay
+on-screen (the explicitness matters: `span-all`’s default alignment centers
+without that shift, so near a corner it would overflow and be rejected).
+Because a fill can only be rejected when its side is shorter than
+`--uktdd-body-min-height` and the two sides sum to the viewport, at least
+one of the pair always fits — for every trigger position, including
+corners, with no height cap needed. The pair is appended via the new
+`--uktdd-body-fill-fallbacks` custom property, so upward-opening dropdowns
+can flip its order to keep the last-resort fill on their preferred side.
+
+In the fill placements the body now spans the trigger (anchor-centered,
+shifted to fit) instead of keeping the primary placement’s edge alignment.
+Consumer fallback lists must stay within the same budget: at most three
+options in `--uktdd-body-position-try-fallbacks`, and a single option in
+`--uktdd-submenu-position-try-fallbacks` (the submenu list is now its
+author fallback, `--uktdd-fill`, `--uktdd-fill flip-inline`, then the two
+block-side fills — also exactly five).
+
+Also fixes the submenu rule’s `--uktdd-body-gap: 0` override: the unitless
+zero made the fill options’ `calc(100% - var(--uktdd-body-gap) * 2)`
+block-size invalid (`<percentage> - <number>`), so every fill option was
+evaluated at natural size and rejected — submenu fill fallbacks never
+applied, and a submenu taller than the space beside its parent item
+overflowed the viewport instead of filling and scrolling. The zero now
+carries a unit (`0px`).
--- a/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
+++ b/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
@@ -17,39 +17,42 @@ were never evaluated: the body opened toward the edge and ran off-screen,
 and consumers had to cap `--uktdd-body-max-height` to work around it.
 
 The four appended tactic fills are replaced by two new named options —
-`--uktdd-fill-bottom` and `--uktdd-fill-top` — bringing the default list to
-exactly five fallbacks, Chromium’s evaluated budget. That is one past the
-spec’s guaranteed floor: an engine at the exact floor would evaluate only
-four and drop the second fill. The design is validated in Chromium — the
-engine whose limit bites — and untested at the floor, since no known engine
-sits there. Each fill pairs a single-side `position-area`
-(`block-end`/`block-start`, spanning all inline tiles so it can never be
-rejected for horizontal overflow) with `justify-self: anchor-center`, which
-centers the body over the trigger and shifts it inward as needed to stay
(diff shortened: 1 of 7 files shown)


packages/dropdown/src/Dropdown.css
```diff
@@ -79,24 +84,27 @@
     /* the primary placement wins whenever the body fits there, then the
        author’s fallbacks are tried in order (no position-try-order:
        most-height, which would override a fitting primary with whichever
```
acusti: i know this is pre-existing, but i just noticed that i don’t understand this clause:
> which would override a fitting primary with whichever side of the viewport happens to be taller

“a fitting primary”? i guess that means “a primary placement that fits”? i actually think this entire comment block needs to be shorter, it’s too much prose to parse as a reader of the code, but that seems like an instance where conciseness compromised comprehensibility.

some other substitutions, all of these in service of concision:

“author’s fallbacks” → “fallbacks”
“when the body fits at its natural size in none” → “when the body fits in none”
“author’s placements” → “placements”
“and lets the content region scroll” → “so the content region scrolls”
“, so a body taller than the space never overflows or covers the trigger.” → “.”
“Engines may stop evaluating position options at the spec’s implementation-defined limit, which can be as low as five (Chromium evaluates exactly five and silently ignores the rest)” → “The limit at which engines may stop evaluating position options can be as low as five (Chromium evaluates five and silently ignores the rest)”
“, so the whole list must land the body on-screen within five options.” → “.”
“is the three named author fallbacks plus these two fills — which is why there are two: one per block side, each unrejectable on the inline axis” → “is the three named author fallbacks plus these two fills: one per block side, each unrejectable on the inline axis”

packages/dropdown/src/Dropdown.css
```diff
@@ -311,27 +325,50 @@ div.uktdropdown-body {
...
+   whichever block side can hold the element is always accepted — that’s
+   what makes the two of them a complete rescue within the five-option
+   evaluation limit (see the .uktdropdown-body rule). A bare block keyword
+   spans all inline tiles, so the containing block is viewport-wide and the
+   element’s max-inline-size can’t overflow it, and an explicit
+   justify-self: anchor-center keeps the element centered over the trigger
+   but shifts it inward as needed to stay inside the viewport. (The
+   explicitness matters: span-all’s default `normal` alignment centers the
+   same way but skips that shift, so near a viewport corner the option
+   would overflow and be rejected.) */
```
acusti: as above, too long to read (the previous version was too long also). look for simplifications similar to the ones i suggested in my previous review comment in this file)

packages/dropdown/README.md
```diff
@@ -604,9 +615,25 @@ The primary placement wins whenever the body fits there at its natural
...
+(`--uktdd-body-fill-fallbacks: --uktdd-fill-top, --uktdd-fill-bottom`) so
+the last-resort fill also prefers the upward side. In these fill placements
+the body spans the trigger (centered over it, shifted inward as needed to
+stay on-screen) rather than keeping the primary placement’s edge alignment.
+
+Keep `--uktdd-body-position-try-fallbacks` to at most three options.
+Engines may stop evaluating `position-try` options at the spec’s
+implementation-defined limit, which can be as low as five — Chromium
+evaluates exactly five and silently ignores the rest — and the component
+appends the two fill placements after your list. A fourth author fallback
```
acusti: let’s replace

> Engines may stop evaluating `position-try` options at the spec’s implementation-defined limit, which can be as low as five — Chromium evaluates exactly five and silently ignores the rest — and the component appends the two fill placements after your list.

with

> The limit at which browsers stop evaluating `position-try` options can be as low as five (e.g. in Chromium) and the component appends the two fill placements after your list.

packages/dropdown/README.md
```diff
@@ -778,7 +805,11 @@ it flush to the parent item’s edge across browsers. Placement custom
 properties follow the established naming scheme:
 
 - `--uktdd-submenu-position-area` (default: `inline-end span-block-end`)
-- `--uktdd-submenu-position-try-fallbacks`
+- `--uktdd-submenu-position-try-fallbacks` (keep it to a single option: the
+  component appends four last-resort placements after it, and engines may
+  evaluate as few as five `position-try` options in total — Chromium
+  evaluates exactly five — so a second author fallback pushes the final
+  rescue placement out of reach)
```
acusti: let’s replace:

> and engines may evaluate as few as five `position-try` options in total — Chromium evaluates exactly five — so a second author fallback pushes the final rescue placement out of reach

with:

> and browsers may evaluate as few as five `position-try` options, e.g. Chromium, so a second author fallback pushes the final rescue placement out of reach

acusti (commented):
this looks great, big improvement on this functionality. would it make sense to add more last-resort fill placements after the authored primary placements for the sake of browsers that support more than 5 options?

packages/dropdown/src/Dropdown.css
acusti: Applied all the substitutions in 5e2bfb5, and rewrote the confusing clause as “which would move the body to the viewport’s taller side even when the primary placement fits”. Also trimmed the closing sentence to “…must keep it to three options so these fills stay within the limit”. The block is down from 22 lines to 17.

---

packages/dropdown/src/Dropdown.css
acusti: Shortened in 5e2bfb5 with the same moves: dropped “author’s”/“available”/“here”, folded the intrinsic-keywords caveat into the natural-size sentence instead of its own parenthetical, and cut the redundant max-block-size sentence (the base clamp applying is the unsurprising case — what matters is min-block-size driving the too-short-side rejection, which stays). 27 lines down to 21.

---

packages/dropdown/README.md
acusti: Replaced verbatim in 5e2bfb5.

---

packages/dropdown/README.md
acusti: Replaced verbatim in 5e2bfb5.

---

acusti:
> would it make sense to add more last-resort fill placements after the authored primary placements for the sake of browsers that support more than 5 options?

I considered it, but extra fills would be dead code in every browser, not just Chromium: options are tried in order and the first that fits wins, and one of `--uktdd-fill-bottom`/`--uktdd-fill-top` always fits (neither can be rejected on the inline axis, and at least one block side is ≥ `--uktdd-body-min-height`). So nothing listed *after* the pair is ever reachable, regardless of the engine's limit.

The variant that would actually benefit generous browsers is inserting higher-fidelity fills *before* the pair — e.g. the old `--uktdd-fill, --uktdd-fill flip-block`, which keep the primary's edge alignment instead of anchor-centering. But that pushes the guaranteed pair to positions 6–7, past Chromium's limit of exactly five, and reintroduces the bug there (I tested that variant: it fixed the edges but still failed at corners, and regressed one corner case the old list handled).

Where a consumer has budget headroom, `--uktdd-body-fill-fallbacks` already supports exactly this: with only 2 authored fallbacks you can set `--uktdd-body-fill-fallbacks: --uktdd-fill, --uktdd-fill-bottom, --uktdd-fill-top` to get an edge-aligned same-side fill first while keeping the guaranteed pair within the limit. Could be worth a README recipe if it comes up in practice.

---

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -0,0 +1,48 @@
...
+Because a fill can only be rejected when its side is shorter than
+`--uktdd-body-min-height` and the two sides sum to the viewport, at least
+one of the pair always fits — for every trigger position, including
+corners, with no height cap needed. The pair is appended via the new
+`--uktdd-body-fill-fallbacks` custom property, so upward-opening dropdowns
+can flip its order to keep the last-resort fill on their preferred side.
+
+In the fill placements the body now spans the trigger (anchor-centered,
+shifted to fit) instead of keeping the primary placement’s edge alignment;
+fills only apply when the body fits nowhere at its natural size. Consumer
```
acusti: i think we can cut this clause (it’s clear from the rest of the explanation): `; fills only apply when the body fits nowhere at its natural size`

acusti (commented):
i forgot to review the changeset first time through. also, can we update the existing dropdown stories to demonstrate these behavior fixes (we can also add a story or two if necessary)?

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Cut (squashed into the main commit, now at head).

---

acusti:
Stories updated in 23e2496:

- New `FillsWhenTooTall` story: the issue-#399 shape — a menu taller than the viewport with triggers pinned to the bottom edge and both corners, no `--uktdd-body-max-height` cap. Each menu flips above via `--uktdd-fill-top`, fills that side, and scrolls; the corner menus demonstrate the anchor-center shift keeping them on-screen; the center menu nests a tall submenu that fills and scrolls beside its parent item (the `0px` gap fix). It renders in its own iframe on the docs page (`inline: false`, 420px) so the fill placements resolve against a real viewport edge rather than the scrolling docs page.
- The two top-opening `DirectionRecipes` now also flip `--uktdd-body-fill-fallbacks`, matching the README’s updated cheatsheet, with a line in the story description noting it.

Verified in the live Storybook in headless Chromium: all three menus open filled/scrolling and fully on-screen, and the submenu fills beside its parent.

---

acusti:
A correction to my claim above that "one of \`--uktdd-fill-bottom\`/\`--uktdd-fill-top\` always fits (… at least one block side is ≥ \`--uktdd-body-min-height\`)" — as originally shipped, that only held while \`--uktdd-body-min-height\` stayed under ~half the viewport.

Each fill is pinned to a single block side of the trigger (\`position-area: block-end\`/\`block-start\` — a single-side area can't span the trigger without covering it), so a fill's vertical room is bounded by where the trigger sits, and a vertically centered trigger caps *both* sides near 50dvb. Above that, the base \`min-block-size: min(--uktdd-body-min-height, --uktdd-body-max-height)\` rejected both fills and the body reverted to the overflowing primary placement. Concretely: \`--uktdd-body-min-height: 300px\`, centered trigger, 420px viewport → ~195px per side, both fills rejected, body renders off-screen (\`bot=634/420\`). It only bit near the center — an edge trigger's roomy side is ~90dvb.

**Resolved outright in 5989e270's ancestor 16137630** rather than documented around: \`@position-try\` accepts sizing properties, so the fills now override \`min-block-size\` to cap the floor at the worst-case larger side, \`(100dvb - anchor-size(block)) / 2\` minus the gap margins. A cramped side is still rejected in favor of the roomier one, but the roomier side can no longer be rejected — the pair is unconditionally accepting for any \`--uktdd-body-min-height\`. Verified in Chromium with the exact repro: the body fills below the centered trigger to the viewport bottom and scrolls, with the computed floor resolving to \`min(300px, 179px) = 179px\`, including the dead-center exact-fit case. 270e5bc9 adds the centered/300px case to the \`FillsWhenTooTall\` story.

(This comment originally pointed at a doc-only caveat commit, since dropped in a branch cleanup once the real fix landed.)

---

packages/docs/stories/Dropdown.stories.tsx
```diff
@@ -725,11 +763,12 @@ export const DirectionRecipes: Story = {
...
                 {DIRECTION_RECIPE_ROWS.map((row) => (
                     <div className="direction-recipes-row" key={row[0].className}>
-                        {row.map(({ className, label }) => (
+                        {row.map(({ className, label, style }) => (
                             <Dropdown
                                 className={className}
                                 key={className}
                                 onSubmitItem={fn()}
+                                style={style}
                             >
```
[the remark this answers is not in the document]
acusti: No change needed — `bun run --filter '@acusti/uikit-docs' tsc` passes clean as written. `readonly` object *properties* (what `as const` produces) are assignable to a mutable-typed target like `CSSProperties & Record<\`--${string}\`, string | number | undefined>`; TypeScript only enforces `readonly` compatibility for arrays and tuples, not object properties, so there's no incompatibility here and no cast is required. And as you note, the `as const` is load-bearing for the `row[0].className` tuple typing, so keeping it is correct.

---

packages/docs/stories/Dropdown.stories.tsx
```diff
@@ -746,6 +785,126 @@ export const DirectionRecipes: Story = {
...
+
+// the recipe inline on the style prop (like DIRECTION_RECIPE_ROWS) so the
+// docs “Show code” panel displays it: the cover fill listed before the
+// guaranteed pair, with two authored fallbacks — the most a three-fill list
+// has room for (six options, the limit every current engine clears)
+const COVER_FILL_STYLE = {
+    '--uktdd-body-fill-fallbacks':
+        '--uktdd-fill-cover, --uktdd-fill-bottom, --uktdd-fill-top',
+    '--uktdd-body-position-try-fallbacks': '--uktdd-top-start, --uktdd-bottom-end',
+} as const;
```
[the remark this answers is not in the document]
acusti: No change needed — `bun run --filter '@acusti/uikit-docs' tsc` passes clean as written. `readonly` object *properties* (what `as const` produces) are assignable to a mutable-typed target like `Dropdown`'s `style` prop (`CSSProperties & Record<\`--${string}\`, string | number | undefined>`); TypeScript only enforces `readonly` compatibility for arrays and tuples, not object properties, so `COVER_FILL_STYLE` is assignable without any cast. (The `as const` here isn't strictly necessary either — the object is assignable without it — but it's harmless, not an error.)

---

.changeset/dropdown-submenu-fills-beside-parent.md
```diff
@@ -0,0 +1,34 @@
+---
+'@acusti/dropdown': minor
+---
+
+Keep a too-tall submenu beside its parent instead of covering the parent
+menu, and give submenus their own last-resort fills decoupled from the
+body’s
```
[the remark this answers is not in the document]
acusti: addressed

PR #404, opened 2026-07-21
Split --uktdd-body-position-try-fallbacks into two fixed slots

acusti:
Stacked on #403.

## Why

The body's author fallbacks sit between the primary placement and the appended fills, and the budget is exactly **two** (2 author fallbacks + 2 fills = 4 past the base = the spec's guaranteed five-option floor). As a comma-separated `--uktdd-body-position-try-fallbacks` list, that limit was only *documented* — a third entry silently pushed a fill past the floor and let a trigger near a viewport edge open off-screen (the #399-class bug this stack exists to prevent).

## What

- **Two fixed slots** — `--uktdd-body-position-try-fallback-1` and `--uktdd-body-position-try-fallback-2` — replace the list, so the two-option budget is *structural*: there's no third slot to overflow into.
- The submenu, which affords a single author fallback, becomes the singular `--uktdd-submenu-position-try-fallback`.
- **New shipped no-op `--uktdd-noop`** for a recipe with one meaningful fallback (a centered menu): set slot 1, leave slot 2 as `--uktdd-noop` so it doesn't inherit the default's corner flip. It's an empty `@position-try` that evaluates as the primary placement — already rejected by the time it's reached — so it's skipped.

Propagated through the direction-recipe table, all examples, the story fixtures/recipe rows, and the comments. Also fixes a lingering three-option list in the FixedHeader avatar-menu fixture (missed in the earlier diagonal-drop pass).

## Migration

Set `--uktdd-body-position-try-fallback-1` / `-2` where you set the two entries of the old list (the README recipe table lists per-direction values), and `--uktdd-submenu-position-try-fallback` (singular) for the submenu. Breaking rename, acceptable at `1.0.0-alpha.6`.

## Verification

Centered-menu story verified in Storybook (Chromium): the split composition `var(--fallback-1), var(--fallback-2), var(fills)` parses and places centered-below correctly, and the `--uktdd-noop` slot-2 doesn't drag it into a corner. Dropdown tests (102), tsc, lint, format all pass.

### DIFF
--- a/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
+++ b/.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
@@ -2,55 +2,47 @@
 '@acusti/dropdown': minor
 ---
 
-Keep the last-resort fill placements within the `position-try` evaluation
+Budget the last-resort fill placements to the `position-try` evaluation
 limit so a too-tall body flips and scrolls instead of opening off-screen
 (fixes #399)
 
-The CSS anchor positioning spec lets engines cap the position options list
-at an implementation-defined length with a floor of five — and the list
-includes the element’s base position, so only four fallbacks past it are
+A too-tall body falls back to last-resort fill placements, which have to
+fit within the `position-try` evaluation limit, or an engine silently drops
+one and the body opens toward a viewport edge instead of flipping and
+scrolling. The CSS anchor positioning spec lets engines cap the position
+options list at an implementation-defined length with a floor of five,
+including the element’s base position, so only four fallbacks past it are
 guaranteed. In practice Chromium is the tightest, evaluating six options
 (five fallbacks past the base) and silently ignoring the rest, while Safari
-and Firefox evaluate many more. The body’s previous list was three author
-fallbacks plus four `--uktdd-fill` tactic variants (seven fallbacks), so
-the block-flipped fills that rescue a near-viewport-height body at a
-block-axis viewport edge were never evaluated: the body opened toward the
-edge and ran off-screen, and consumers had to cap `--uktdd-body-max-height`
-to work around it.
+and Firefox evaluate many more.
 
-The four appended tactic fills are replaced by two new named options —
-`--uktdd-fill-bottom` and `--uktdd-fill-top` — and the default author
-fallback list drops from three options to two (the block flip and inline
-flip of the primary; the both-axes diagonal is dropped). The full list is
-now two author fallbacks plus the two fills — four fallbacks past the base
-position, exactly the spec’s guaranteed floor of five options, so no engine
-can silently drop a fill. The cost of dropping the diagonal is that a
-trigger crammed into the corner diagonal to the primary opening direction
-lands in a fill rather than a fourth natural placement, which for a
-corner-pinned trigger is a re-tune-the-primary case anyway. Each fill pairs
-a single-side `position-area` (`block-end`/`block-start`, spanning all
-inline tiles so it can never be rejected for horizontal overflow) with
+The default budget is two author fallbacks plus two named fill options,
+`--uktdd-fill-bottom` and `--uktdd-fill-top`, so no engine can silently
+drop a fill. The two author fallbacks are the single-axis flips of the
+primary (a block flip and an inline flip); the opposite diagonal flip is
+dropped to stay within the floor. The cost of dropping the diagonal is that
+a trigger crammed into the corner diagonal to the primary opening direction
+lands in a fill rather than a fourth natural placement. Each fill pairs a
+single-side `position-area` (`block-end`/`block-start`, spanning all inline
+tiles so it can never be rejected for horizontal overflow) with
 `justify-self: anchor-center`, which centers the body over the trigger and
 shifts it inward as needed to stay on-screen (the explicitness matters:
 `span-all`’s default alignment centers without that shift, so near a corner
 it would overflow and be rejected). A fill is only rejected when its side
-is shorter than the body’s min-block-size floor — what sends a cramped
-side’s fill to the roomier opposite side — and each fill caps that floor at
-the worst-case larger side (half the viewport minus the trigger, the most a
+is shorter than the body’s `min-block-size` floor, and each fill caps that
+floor at the larger side (roughly half the viewport, the most a
 block-centered trigger can guarantee), so at least one of the pair always
-fits — for every trigger position, including corners and a
-`--uktdd-body-min-height` above half the viewport, with no height cap
-needed. The pair is appended via the new `--uktdd-body-fill-fallbacks`
-custom property, so upward-opening dropdowns can flip its order to keep the
-last-resort fill on their preferred side. (Submenus have their own
-last-resort fills, `--uktdd-submenu-fill-fallbacks`, rather than reusing
-this pair — see the separate submenu changeset.)
+fits, no additional height cap needed. The pair is appended via the new
+`--uktdd-body-fill-fallbacks` custom property, so upward-opening dropdowns
+can flip its order to keep the last-resort fill on their preferred side.
+(Submenus have their own last-resort fills,
+`--uktdd-submenu-fill-fallbacks`, rather than reusing this pair — see the
+separate submenu changeset.)
 
-In the fill placements the body now spans the trigger (anchor-centered,
-shifted to fit) instead of keeping the primary placement’s edge alignment.
-Consumer fallback lists must stay within the budget: at most two options in
-`--uktdd-body-position-try-fallbacks` (matching the new floor-safe
-default), and a single option in `--uktdd-submenu-position-try-fallbacks`.
+In the fill placements the body spans the trigger instead of keeping the
+primary placement’s edge alignment. Consumer overrides stay within the
+budget: the two `--uktdd-body-position-try-fallback-1`/`-2` slots plus the
+two fills, and the single `--uktdd-submenu-position-try-fallback`.
 
 Also fixes the submenu rule’s `--uktdd-body-gap: 0` override: the unitless
 zero made the fill options’ `calc(100% - var(--uktdd-body-gap) * 2)`
--- a/.changeset/dropdown-rename-position-try-fallbacks.md
+++ b/.changeset/dropdown-rename-position-try-fallbacks.md
@@ -28,7 +28,7 @@ physical keyword on one axis with a logical one on the other — and read as
 `top`/`bottom` in the near-universal horizontal-tb writing mode.
 
 **Migration:** if you referenced any of the renamed `@position-try` block
-names in `--uktdd-body-position-try-fallbacks`, or set
+names in `--uktdd-body-position-try-fallback-1`/`-2`, or set
 `--uktdd-body-position-area`/`--uktdd-submenu-position-area` directly with
 `top`/`bottom`/`span-left`/`span-right`, update them:
 
@@ -37,14 +37,14 @@ names in `--uktdd-body-position-try-fallbacks`, or set
 .my-dropdown {
     --uktdd-body-position-area: bottom span-left;
     --uktdd-body-position-try-fallbacks:
-        --uktdd-top-right, --uktdd-bottom-left, --uktdd-top-left;
+        --uktdd-top-right, --uktdd-bottom-left;
 }
 
-/* after */
+/* after (the fallback list is now two fixed slots) */
 .my-dropdown {
     --uktdd-body-position-area: block-end span-inline-start;
-    --uktdd-body-position-try-fallbacks:
-        --uktdd-top-end, --uktdd-bottom-start, --uktdd-top-start;
+    --uktdd-body-position-try-fallback-1: --uktdd-top-end;
+    --uktdd-body-position-try-fallback-2: --uktdd-bottom-start;
 }
 ```
 
--- a/.changeset/dropdown-split-fallback-slots.md
+++ b/.changeset/dropdown-split-fallback-slots.md
@@ -1,0 +1,30 @@
+---
+'@acusti/dropdown': major
+---
+
+Replace the `--uktdd-body-position-try-fallbacks` list with two fixed
+slots, `--uktdd-body-position-try-fallback-1` and
+`--uktdd-body-position-try-fallback-2`, and rename the submenu’s list to
+the singular `--uktdd-submenu-position-try-fallback`
+
+The author fallbacks sit between the primary placement and the appended
+fills, and the component’s budget is exactly two of them (two author
+fallbacks + two fills = four, plus the base = the spec’s guaranteed
+five-option floor). As a comma-separated list, that limit was only
+documented, and a third entry silently pushed a fill past the floor,
+letting a trigger near a viewport edge open off-screen. Splitting the list
+into two named slots makes the two-option budget structural: there is no
+third slot to overflow into. The submenu, which affords a single author
+fallback, likewise becomes a single named slot.
+
+Migration: set `--uktdd-body-position-try-fallback-1` and `-2` where you
+previously set the two entries of `--uktdd-body-position-try-fallbacks`
+(the direction-recipe table in the README lists the per-direction values),
+and `--uktdd-submenu-position-try-fallback` (singular) where you set the
+submenu list.
+
+Also adds a shipped no-op placement, `--uktdd-noop`, for a recipe with a
+single meaningful fallback (a centered menu, say): set slot 1 and leave
+slot 2 as `--uktdd-noop` so it doesn’t inherit the default’s flip. It
+applies no overrides, so it evaluates as the primary placement and is
+skipped.
--- a/packages/docs/stories/Dropdown.stories.tsx
+++ b/packages/docs/stories/Dropdown.stories.tsx
@@ -712,8 +712,8 @@ const DIRECTION_RECIPE_ROWS = [
             label: 'Bottom Start (default)',
             style: {
                 '--uktdd-body-position-area': 'block-end span-inline-end',
-                '--uktdd-body-position-try-fallbacks':
-                    '--uktdd-top-start, --uktdd-bottom-end',
+                '--uktdd-body-position-try-fallback-1': '--uktdd-top-start',
+                '--uktdd-body-position-try-fallback-2': '--uktdd-bottom-end',
             },
         },
         {
@@ -721,8 +721,8 @@ const DIRECTION_RECIPE_ROWS = [
             label: 'Bottom End',
             style: {
                 '--uktdd-body-position-area': 'block-end span-inline-start',
-                '--uktdd-body-position-try-fallbacks':
-                    '--uktdd-top-end, --uktdd-bottom-start',
+                '--uktdd-body-position-try-fallback-1': '--uktdd-top-end',
+                '--uktdd-body-position-try-fallback-2': '--uktdd-bottom-start',
             },
         },
     ],
@@ -733,8 +733,8 @@ const DIRECTION_RECIPE_ROWS = [
             style: {
                 '--uktdd-body-fill-fallbacks': '--uktdd-fill-top, --uktdd-fill-bottom',
                 '--uktdd-body-position-area': 'block-start span-inline-end',
-                '--uktdd-body-position-try-fallbacks':
-                    '--uktdd-bottom-start, --uktdd-top-end',
+                '--uktdd-body-position-try-fallback-1': '--uktdd-bottom-start',
+                '--uktdd-body-position-try-fallback-2': '--uktdd-top-end',
             },
         },
         {
@@ -743,8 +743,8 @@ const DIRECTION_RECIPE_ROWS = [
             style: {
                 '--uktdd-body-fill-fallbacks': '--uktdd-fill-top, --uktdd-fill-bottom',
                 '--uktdd-body-position-area': 'block-start span-inline-start',
-                '--uktdd-body-position-try-fallbacks':
-                    '--uktdd-bottom-end, --uktdd-top-start',
+                '--uktdd-body-position-try-fallback-1': '--uktdd-bottom-end',
+                '--uktdd-body-position-try-fallback-2': '--uktdd-top-start',
             },
         },
     ],
@@ -754,7 +754,7 @@ export const DirectionRecipes: Story = {
     parameters: {
         docs: {
             description: {
-                story: 'The four `@position-try` recipes the component ships, each pairing `--uktdd-body-position-area` with its matching `--uktdd-body-position-try-fallbacks` (see the README’s “Changing the Default Direction” section for the full cheatsheet). Each dropdown carries its recipe on its `style` prop, so “Show code” below displays the exact variable pairs. Each is named for the edge that stays flush with the trigger, not the direction the body extends toward: “start” keeps the body’s inline-start edge flush with the trigger’s, extending toward inline-end; “end” is the mirror image. The top-opening pair also flips `--uktdd-body-fill-fallbacks` so the last-resort fill prefers the upward side. Each recipe opens in its named direction whenever that side has room, independent of where the trigger sits; the bottom pair is placed near the top of the canvas and the top pair near the bottom only so each menu opens into empty space instead of overlapping the other row.',
+                story: 'The four `@position-try` recipes the component ships, each pairing `--uktdd-body-position-area` with its matching `--uktdd-body-position-try-fallback-1` and `-2` (see the README’s [“Changing the Default Direction”](https://github.com/acusti/uikit/tree/main/packages/dropdown#changing-the-default-direction) section for the full cheatsheet). Each dropdown carries its recipe on its `style` prop, so “Show code” below displays the exact variables. Each is named for the edge that stays flush with the trigger, not the direction the body extends toward: “start” keeps the body’s inline-start edge flush with the trigger’s, extending toward inline-end; “end” is the mirror image. The top-opening pair also flips `--uktdd-body-fill-fallbacks` so the last-resort fill prefers the upward side. Each recipe opens in its named direction whenever that side has room, independent of where the trigger sits; the bottom pair is placed near the top of the canvas and the top pair near the bottom only so each menu opens into empty space instead of overlapping the other row.',
             },
         },
     },
@@ -842,7 +842,8 @@ export const FillsWhenTooTall: Story = {
 const COVER_FILL_STYLE = {
     '--uktdd-body-fill-fallbacks':
         '--uktdd-fill-cover, --uktdd-fill-bottom, --uktdd-fill-top',
-    '--uktdd-body-position-try-fallbacks': '--uktdd-top-start, --uktdd-bottom-end',
+    '--uktdd-body-position-try-fallback-1': '--uktdd-top-start',
+    '--uktdd-body-position-try-fallback-2': '--uktdd-bottom-end',
 } as const;
 
 export const CoverFillRecipe: Story = {
--- a/packages/dropdown/README.md
+++ b/packages/dropdown/README.md
@@ -524,7 +524,9 @@ Placement is best customized in CSS instead of props. The component exposes
 CSS custom properties for the most common low-level placement controls:
 
 - `--uktdd-body-position-area`
-- `--uktdd-body-position-try-fallbacks`
+- `--uktdd-body-position-try-fallback-1` and
+  `--uktdd-body-position-try-fallback-2` (the two author fallbacks, tried
+  in order after the primary placement)
 - `--uktdd-body-fill-fallbacks` (the last-resort fill placements, tried
   when the body fits nowhere at its natural size)
 - `--uktdd-body-gap` (space between the trigger and the body)
@@ -538,8 +540,8 @@ Example:
 ```css
 .settings-dropdown {
     --uktdd-body-position-area: block-end span-inline-start;
-    --uktdd-body-position-try-fallbacks:
-        --uktdd-top-end, --uktdd-bottom-start;
+    --uktdd-body-position-try-fallback-1: --uktdd-top-end;
+    --uktdd-body-position-try-fallback-2: --uktdd-bottom-start;
     --uktdd-body-gap: 8px;
 }
 
@@ -553,15 +555,17 @@ placement and sizing control when a product surface needs it.
 
 ### Changing the Default Direction
 
-`--uktdd-body-position-area` and `--uktdd-body-position-try-fallbacks`
-always change together. `--uktdd-body-position-area` is the primary
-placement, used whenever it fits in the viewport — that’s what actually
-determines the direction a dropdown opens by default. The fallbacks list is
-only consulted when the primary placement doesn’t fit, so overriding it
-alone changes nothing in the common case where the primary already fits.
-Overriding the primary alone works, but leaves behind a fallback cascade
-tuned for the old primary — in a cramped viewport, the dropdown can fall
-back toward the direction you just moved away from.
+`--uktdd-body-position-area` and its two fallback slots
+(`--uktdd-body-position-try-fallback-1`,
+`--uktdd-body-position-try-fallback-2`) always change together.
+`--uktdd-body-position-area` is the primary placement, used whenever it
+fits in the viewport — that’s what actually determines the direction a
+dropdown opens by default. The fallback slots are tried in order only when
+the primary placement doesn’t fit. Set all three together to ensure the
+menu behaves the way you want it to in all possible contexts. If the new
+direction has only one meaningful fallback (a centered menu, say), put it
+in slot 1 and set slot 2 to the shipped no-op `--uktdd-noop` so it doesn’t
+inherit that flip.
 
 The four `@position-try` blocks the component ships (`--uktdd-top-start`,
 `--uktdd-top-end`, `--uktdd-bottom-start`, `--uktdd-bottom-end`) are named
@@ -576,15 +580,18 @@ physical keyword on one axis with a logical one on the other — though
 `block-start`/`block-end` read as `top`/`bottom` in the near-universal
 horizontal-tb writing mode.
 
-Pick the row matching the direction you want, and set both variables to its
-pair:
+Pick the row matching the direction you want, and set the three CSS custom
+properties to the corresponding values:
 
-| Direction                           | `--uktdd-body-position-area`    | `--uktdd-body-position-try-fallbacks`   |
-| ----------------------------------- | ------------------------------- | --------------------------------------- |
-| Bottom, start-aligned (the default) | `block-end span-inline-end`     | `--uktdd-top-start, --uktdd-bottom-end` |
-| Bottom, end-aligned                 | `block-end span-inline-start`   | `--uktdd-top-end, --uktdd-bottom-start` |
-| Top, start-aligned                  | `block-start span-inline-end`   | `--uktdd-bottom-start, --uktdd-top-end` |
-| Top, end-aligned                    | `block-start span-inline-start` | `--uktdd-bottom-end, --uktdd-top-start` |
+| Direction                           | `--uktdd-body-position-area`    | `…try-fallback-1`      | `…try-fallback-2`      |
+| ----------------------------------- | ------------------------------- | ---------------------- | ---------------------- |
+| Bottom, start-aligned (the default) | `block-end span-inline-end`     | `--uktdd-top-start`    | `--uktdd-bottom-end`   |
+| Bottom, end-aligned                 | `block-end span-inline-start`   | `--uktdd-top-end`      | `--uktdd-bottom-start` |
+| Top, start-aligned                  | `block-start span-inline-end`   | `--uktdd-bottom-start` | `--uktdd-top-end`      |
+| Top, end-aligned                    | `block-start span-inline-start` | `--uktdd-bottom-end`   | `--uktdd-top-start`    |
+
+(The `…try-fallback-N` columns are `--uktdd-body-position-try-fallback-1`
+and `-2`.)
 
 The two Top rows should also flip the last-resort fill pair to match:
 `--uktdd-body-fill-fallbacks: --uktdd-fill-top, --uktdd-fill-bottom`.
@@ -595,8 +602,8 @@ trigger pinned to the bottom of the viewport, near the inline-end edge):
 ```css
 .bottom-toolbar-dropdown {
     --uktdd-body-position-area: block-start span-inline-start;
-    --uktdd-body-position-try-fallbacks:
-        --uktdd-bottom-end, --uktdd-top-start;
+    --uktdd-body-position-try-fallback-1: --uktdd-bottom-end;
+    --uktdd-body-position-try-fallback-2: --uktdd-top-start;
     --uktdd-body-fill-fallbacks: --uktdd-fill-top, --uktdd-fill-bottom;
 }
 ```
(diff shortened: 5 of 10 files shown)


.changeset/dropdown-split-fallback-slots.md
```diff
@@ -0,0 +1,30 @@
+---
+'@acusti/dropdown': minor
```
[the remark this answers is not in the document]
acusti: Agreed — bumped to `major` in 1f247e58. It's a breaking rename of public custom properties, and it matches the repo's convention (e.g. `dropdown-rename-position-try-fallbacks` is major for renaming this same property). The dropdown alpha train is already headed to a major release, so this just categorizes the change correctly under Major Changes.

.changeset/dropdown-split-fallback-slots.md
```diff
@@ -0,0 +1,30 @@
...
+---
+
+Replace the `--uktdd-body-position-try-fallbacks` list with two fixed
+slots, `--uktdd-body-position-try-fallback-1` and
+`--uktdd-body-position-try-fallback-2`, and rename the submenu’s list to
+the singular `--uktdd-submenu-position-try-fallback`
```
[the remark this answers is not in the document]
acusti: Good catch — reconciled in 1f247e58. Updated the sibling changesets that referenced the old list-form names to the slot names so the combined release notes are consistent: `dropdown-fill-fallbacks-within-evaluation-limit`, `dropdown-submenu-fills-beside-parent`, `dropdown-rename-position-try-fallbacks` (its "after" example is now the two-slot form), and `nested-dropdown-submenus` (→ singular `--uktdd-submenu-position-try-fallback`).

One I left for the maintainer: `dropdown-prefer-authored-direction` still references the old names *and* describes the superseded "four appended `--uktdd-fill` fallbacks" design (this session's fill work replaced those four tactic variants with two named fills). That's a design-narrative change beyond a name swap, so it wants an editorial decision rather than a silent rewrite.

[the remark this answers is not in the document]
acusti: Update: reconciled `dropdown-prefer-authored-direction` too (0c7ed9ff), and reframed the fill narrative across the train. Since the whole fill system churned within this one unreleased alpha — `prefer-authored-direction` introduced four `--uktdd-fill` tactic fallbacks, the evaluation-limit change then replaced them with two named fills — describing "added four, then replaced with two" only muddied the notes. Now `prefer-authored-direction` keeps its actual contribution (authored direction wins over most-height, generic fill-and-scroll last resort) with no four-fill mechanics, and the evaluation-limit changeset describes the two named fills as the design rather than a replacement. The combined changelog now reads as one coherent final-state story.

packages/dropdown/src/Dropdown.css
```diff
@@ -18,7 +18,16 @@
...
     --uktdd-body-position-area: block-end span-inline-end;
-    --uktdd-body-position-try-fallbacks: --uktdd-top-start, --uktdd-bottom-end;
+    /* the two author fallbacks, tried in order after the primary placement
+       and before the fills — one slot each rather than a list so the
+       two-option budget is structural (a fill dropped past the evaluation
+       limit opens a body off-screen; see the .uktdropdown-body rule). The
+       defaults are the primary’s single-axis flips: -1 the block flip, -2
+       the inline flip. Change direction by setting both together (see the
+       recipes in the README); the both-axes diagonal is intentionally not a
+       slot, so a trigger in that corner lands in a fill */
```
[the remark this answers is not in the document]
acusti: Fixed in b80d0fe9 — the comment now reads "Change direction by resetting `--uktdd-body-position-area` and both slots together … the primary determines the direction, the slots are its flips," matching the README's "set all three together" guidance.

packages/dropdown/README.md
```diff
@@ -669,8 +681,8 @@ rather than one side of it.
 
 ```css
 .tall-menu {
-    --uktdd-body-position-try-fallbacks:
-        --uktdd-top-start, --uktdd-bottom-end;
+    --uktdd-body-position-try-fallback-1: --uktdd-top-start;
+    --uktdd-body-position-try-fallback-2: --uktdd-bottom-end;
     --uktdd-body-fill-fallbacks:
         --uktdd-fill-cover, --uktdd-fill-bottom, --uktdd-fill-top;
```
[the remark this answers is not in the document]
acusti: There's no contradiction here — the "five" in the PR/changesets is the **spec floor** (five *options* = the base position + four fallbacks), not Chromium's limit. Chromium evaluates exactly **six** options (base + five fallbacks); this was measured directly with a size-based-rejection test harness, and Safari 26.5 and Firefox 152 both evaluate well past twelve. The fill-fallbacks changeset states it explicitly: "evaluating six options (five fallbacks past the base)."

So the guidance is accurate and already balanced: the default list stays within the spec-guaranteed five-option floor, and the opt-in cover recipe (six options) is one above that floor — safe on every current engine, with the section immediately noting you can drop to one author fallback to keep the whole list within the spec floor. Removing the empirical per-engine data would undo a deliberate de-hedge made after that cross-engine testing.

packages/docs/stories/Dropdown.stories.tsx
```diff
@@ -842,7 +842,8 @@ export const FillsWhenTooTall: Story = {
 const COVER_FILL_STYLE = {
     '--uktdd-body-fill-fallbacks':
         '--uktdd-fill-cover, --uktdd-fill-bottom, --uktdd-fill-top',
-    '--uktdd-body-position-try-fallbacks': '--uktdd-top-start, --uktdd-bottom-end',
+    '--uktdd-body-position-try-fallback-1': '--uktdd-top-start',
```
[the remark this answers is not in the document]
acusti: Same as the README thread: the story description already frames both sides. "Six options, within the limit every current engine evaluates" is a measured fact (Chromium exactly six, Safari 26.5 / Firefox 152 both >12), and cover is explicitly opt-in — the README section and changeset both note dropping to one author fallback to stay within the spec-guaranteed floor of five. The strict-limit motivation is about the *default* (five options, floor-safe); the cover recipe is a deliberate one-above-the-floor opt-in, not a claim that six is universally guaranteed.

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -2,44 +2,41 @@
...
-Keep the last-resort fill placements within the `position-try` evaluation
+Budget the last-resort fill placements to the `position-try` evaluation
 limit so a too-tall body flips and scrolls instead of opening off-screen
 (fixes #399)
 
-The CSS anchor positioning spec lets engines cap the position options list
-at an implementation-defined length with a floor of five — and the list
+The last-resort fills a too-tall body falls back to have to fit within the
+`position-try` evaluation limit, or an engine silently drops one and the
+body opens toward a viewport edge instead of flipping and scrolling. The
```
acusti: i don’t follow the beginning of this sentence: “The last-resort fills a too-tall body falls back…”

seems like too many verbs? “fills” and “falls back”?

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -2,44 +2,41 @@
...
+Each fill pairs a single-side `position-area` (`block-end`/`block-start`,
+spanning all inline tiles so it can never be rejected for horizontal
+overflow) with `justify-self: anchor-center`, which centers the body over
+the trigger and shifts it inward as needed to stay on-screen (the
+explicitness matters: `span-all`’s default alignment centers without that
+shift, so near a corner it would overflow and be rejected). A fill is only
+rejected when its side is shorter than the body’s min-block-size floor —
+what sends a cramped side’s fill to the roomier opposite side — and each
+fill caps that floor at the worst-case larger side (half the viewport minus
+the trigger, the most a block-centered trigger can guarantee), so at least
```
acusti: technically our math doesn’t include the trigger, right? just the gap?

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -48,9 +45,9 @@ this pair — see the separate submenu changeset.)
 
 In the fill placements the body now spans the trigger (anchor-centered,
 shifted to fit) instead of keeping the primary placement’s edge alignment.
```
acusti: is this somewhat obviated by this explanation above?
> …with `justify-self: anchor-center`, which centers the body over the trigger and shifts it inward as needed to stay on-screen

packages/docs/stories/Dropdown.stories.tsx
```diff
@@ -754,7 +754,7 @@ export const DirectionRecipes: Story = {
...
+                story: 'The four `@position-try` recipes the component ships, each pairing `--uktdd-body-position-area` with its matching `--uktdd-body-position-try-fallback-1` and `-2` (see the README’s “Changing the Default Direction” section for the full cheatsheet). Each dropdown carries its recipe on its `style` prop, so “Show code” below displays the exact variables. Each is named for the edge that stays flush with the trigger, not the direction the body extends toward: “start” keeps the body’s inline-start edge flush with the trigger’s, extending toward inline-end; “end” is the mirror image. The top-opening pair also flips `--uktdd-body-fill-fallbacks` so the last-resort fill prefers the upward side. Each recipe opens in its named direction whenever that side has room, independent of where the trigger sits; the bottom pair is placed near the top of the canvas and the top pair near the bottom only so each menu opens into empty space instead of overlapping the other row.',
```
acusti: can we link `“Changing the Default Direction” section` to `https://github.com/acusti/uikit/tree/main/packages/dropdown#changing-the-default-direction`

packages/dropdown/src/Dropdown.css
```diff
@@ -18,7 +18,18 @@
...
+    /* the two author fallbacks, tried in order after the primary placement
+       and before the fills — one slot each rather than a list so the
+       two-option budget is structural (a fill dropped past the evaluation
+       limit opens a body off-screen; see the .uktdropdown-body rule). The
+       defaults are the primary’s single-axis flips: -1 the block flip, -2
+       the inline flip. Change direction by resetting --uktdd-body-position-area
+       and both slots together (see the recipes in the README) — the primary
+       determines the direction, the slots are its flips; the both-axes
+       diagonal is intentionally not a slot, so a trigger in that corner
+       lands in a fill */
```
acusti: this is too long. maybe we don’t need to mention “one slot each rather than a list so the two-option budget is structural”? and i think instead of “-1” and “-2”, we can just refer to them as “1” and “2”. probably other things we could tighten. the issue is that it’s intimidating as a code reader to try to wade through, so my brain turns off when i see that giant block of text in a CSS file.

packages/dropdown/src/Dropdown.css
```diff
@@ -35,7 +46,10 @@
     --uktdd-body-color-path: currentColor;
     --uktdd-label-pad-right: 0.625em;
     --uktdd-submenu-position-area: inline-end span-block-end;
-    --uktdd-submenu-position-try-fallbacks: --uktdd-submenu-inline-start;
+    /* the submenu’s single author fallback (the natural opposite-side
+       placement). One slot, not a list: the submenu budget affords only one
+       author fallback before the two fills reach the evaluation limit */
```
acusti: maybe just:
```suggestion
    /* the submenu’s only author fallback (the natural opposite placement) */
```

packages/dropdown/src/Dropdown.css
```diff
@@ -114,10 +128,12 @@
        rejected when its side is shorter than its min-block-size floor,
        and the fills cap that floor at the worst-case larger side (see
        their @position-try rules), so at least one of the pair always
-       fits. Authors overriding --uktdd-body-position-try-fallbacks must
-       keep it to two options so the fills stay within the floor */
+       fits. The two author fallbacks are separate custom properties, not
+       a list, so the two-slot budget is structural — there’s no third
+       slot to overflow the floor into */
```
acusti: i think we can cut the whole last sentence:
```suggestion
       fits */
```

packages/dropdown/src/Dropdown.css
```diff
@@ -237,19 +253,26 @@
...
+           block sides sum to the viewport and one always fits — they carry
+           no at-least-one-always-fits guarantee; they assume at least one
+           inline side has room for the submenu’s width. A submenu both too
+           tall to open beside its parent and too wide for either inline side
+           (a parent menu spanning most of the viewport) exhausts the fills
+           and reverts to its primary placement, which can overflow. That
+           case is out of scope for the narrow columns of a macOS-style menu;
+           a consumer who needs it can append the block-cover fills
+           (--uktdd-fill-bottom, --uktdd-fill-top) to
            --uktdd-submenu-fill-fallbacks */
```
acusti: yowzer, talk about big walls of text in CSS files. one easy ting: no need to talk about what the code doesn’t do, e.g. “(the block-cover pair is gone, so no engine reliance)”, so we can cut anything that spends time on that. full disclosure: i didn’t read the whole thing, it’s too long for me right now at the end of the day, but maybe you can see other ways to make it more approachable?

packages/dropdown/src/Dropdown.css
```diff
@@ -315,6 +338,16 @@ div.uktdropdown-body {
         var(--uktdd-body-pad-bottom) var(--uktdd-body-pad-left);
 }
 
+/* An empty option for a fallback slot a recipe doesn’t use — e.g. a
+   centered menu (bottom/top span-all) has one meaningful author fallback,
+   so it sets --uktdd-body-position-try-fallback-1 and leaves slot 2 as
+   this no-op. It applies no overrides, so it evaluates identically to the
+   primary placement; reached only after the primary was already rejected,
+   it is rejected in turn and skipped — never selected, never breaking the
+   declaration the way an empty var() would. */
```
acusti: maybe:
```suggestion
/* An empty option for a fallback slot a recipe doesn’t use — e.g. a
   centered menu (bottom/top span-all) has one meaningful author fallback,
   so it sets --uktdd-body-position-try-fallback-1 and leaves slot 2 as
   --uktdd-noop, which applies no overrides and evaluates identically to the
   primary placement, meaning it’s rejected in turn and skipped without breaking
   the declaration the way an empty var() would. */
```

packages/dropdown/README.md
```diff
@@ -553,15 +555,21 @@ placement and sizing control when a product surface needs it.
...
+dropdown opens by default. The fallback slots are only consulted when the
+primary placement doesn’t fit, so overriding them alone changes nothing in
+the common case where the primary already fits. Overriding the primary
+alone works, but leaves behind fallback slots tuned for the old primary —
+in a cramped viewport, the dropdown can fall back toward the direction you
+just moved away from — so set all three together, and update both slots:
+leaving one at its default inherits a flip tuned for the old primary, the
+same footgun on a smaller scale. If the new direction has only one
+meaningful fallback (a centered menu, say), put it in slot 1 and set slot 2
+to the shipped no-op `--uktdd-noop` so it doesn’t inherit that flip.
```
acusti: how about:
```suggestion
just moved away from — so set all three together, and update both slots.
If the new direction has only one meaningful fallback (a centered menu,
say), put it in slot 1 and set slot 2 to the shipped no-op `--uktdd-noop`
so it doesn’t inherit that flip.
```

packages/dropdown/README.md
```diff
@@ -639,22 +650,23 @@ trigger can guarantee — so a min-height above half the viewport (honored in
...
-above) for triggers pinned to a viewport corner.
+There are exactly two author fallback slots by design. The spec lets
+engines cap the position options list at an implementation-defined length
+with a floor of five — and that list includes the element’s base position,
+so only four fallbacks past it are guaranteed. The component appends the
+two fill placements after your two slots, which is exactly four — within
+the floor on every conformant engine. That’s why the fallbacks are two
+separate custom properties rather than a list: a list invites a third
+entry, which would push a fill past the floor and let a trigger near a
+viewport edge open off-screen instead of flipping and scrolling. The two
```
acusti: maybe (will require re-formatting the file):
```suggestion
separate custom properties rather than a list: to avoid a third (or more)
entry, which could result in a trigger near a viewport edge opening
off-screen instead of flipping and scrolling. The two
```

packages/dropdown/README.md
```diff
@@ -859,11 +875,10 @@ it flush to the parent item’s edge across browsers. Placement custom
...
 - `--uktdd-submenu-position-area` (default: `inline-end span-block-end`)
-- `--uktdd-submenu-position-try-fallbacks` (default:
-  `--uktdd-submenu-inline-start`; keep it to a single option — the
-  component appends two last-resort fills after it, and the spec guarantees
-  only four evaluated fallbacks, so a second author fallback would push a
-  fill out of reach)
+- `--uktdd-submenu-position-try-fallback` (default:
+  `--uktdd-submenu-inline-start`; a single slot, not a list — the component
+  appends two last-resort fills after it, and the spec guarantees only four
+  evaluated fallbacks, so there’s only room for the one author fallback)
```
acusti: let’s just say
```suggestion
- `--uktdd-submenu-position-try-fallback` (default:
  `--uktdd-submenu-inline-start`; a single slot, not a list)
```

packages/dropdown/README.md
```diff
@@ -875,9 +890,9 @@ properties follow the established naming scheme:
   out of scope for a macOS-style menu’s narrow columns — is the one case
   this gives up versus covering the parent; append
   `--uktdd-fill-bottom, --uktdd-fill-top` here to restore the
-  parent-covering rescue if you need it, keeping
-  `--uktdd-submenu-position-try-fallbacks` empty to stay within the
-  evaluation limit)
+  parent-covering rescue if you need it, leaving
+  `--uktdd-submenu-position-try-fallback` at its default single option to
+  stay within the evaluation limit)
```
acusti: how about:
```suggestion
default. A submenu wider than the space on either side of its parent is
the one case this default doesn’t handle; append
`--uktdd-fill-bottom, --uktdd-fill-top` here to restore the
  parent-covering rescue if you need it
```

acusti (commented):
this behavior is all complicated enough that if there is too much text, even if it’s all very accurate, it becomes overwhelming for a human reader, so i suggested some ways to try to simplify explanations and prevent that overwhelm

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Reworded in e5a75846 — now "A too-tall body falls back to last-resort fill placements, which have to fit within the `position-try` evaluation limit…" (drops the "fills … falls back to" verb pileup).

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Good question, but the trigger *is* in it: the cap is `(100dvb - anchor-size(block)) / 2 - gap·2` — so it subtracts both the trigger (anchor-size) and the gap. The changeset actually omitted the gap. Rather than spell out the full formula in release notes, I simplified it to "roughly half the viewport" (e5a75846); the precise `(100dvb - anchor)/2` lives in the @position-try comment.

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Yes — cut the redundant "(anchor-centered, shifted to fit)" in e5a75846; the sentence now just makes the distinct point (fills span the trigger instead of keeping the primary's edge alignment).

packages/docs/stories/Dropdown.stories.tsx
acusti: Linked in e5a75846 → https://github.com/acusti/uikit/tree/main/packages/dropdown#changing-the-default-direction

packages/dropdown/src/Dropdown.css
acusti: Tightened in e5a75846 — dropped the structural-budget clause (it's covered in the .uktdropdown-body rule) and the diagonal note, and switched to "slot 1 / slot 2". Down to four lines.

packages/dropdown/src/Dropdown.css
acusti: Applied your one-liner in e5a75846.

packages/dropdown/src/Dropdown.css
acusti: Cut in e5a75846 — ends at "…always fits".

packages/dropdown/src/Dropdown.css
acusti: Reworked in e5a75846: dropped the "what it doesn't do" bits (the block-cover-pair / no-engine-reliance line) and the floor arithmetic, and condensed the no-guarantee caveat. About half the length now.

packages/dropdown/src/Dropdown.css
acusti: Applied your suggestion in e5a75846.

packages/dropdown/README.md
acusti: Applied your suggestion in e5a75846.

packages/dropdown/README.md
acusti: Applied your suggestion in e5a75846 (reformatted).

packages/dropdown/README.md
acusti: Applied your suggestion in e5a75846.

packages/dropdown/README.md
acusti: Applied your suggestion in e5a75846.

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -2,55 +2,51 @@
...
-block-axis viewport edge were never evaluated: the body opened toward the
-edge and ran off-screen, and consumers had to cap `--uktdd-body-max-height`
-to work around it.
+A too-tall body falls back to last-resort fill placements, which have to
+fit within the `position-try` evaluation limit, or an engine silently drops
+one and the body opens toward a viewport edge instead of flipping and
+scrolling. The CSS anchor positioning spec lets engines cap the position
+options list at an implementation-defined length with a floor of five — and
+the list includes the element’s base position, so only four fallbacks past
+it are guaranteed. In practice Chromium is the tightest, evaluating six
```
acusti:
```suggestion
options list at an implementation-defined length with a floor of five,
including the element’s base position, so only four fallbacks past
```

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -2,55 +2,51 @@
...
-shifts it inward as needed to stay on-screen (the explicitness matters:
-`span-all`’s default alignment centers without that shift, so near a corner
-it would overflow and be rejected). A fill is only rejected when its side
-is shorter than the body’s min-block-size floor — what sends a cramped
-side’s fill to the roomier opposite side — and each fill caps that floor at
-the worst-case larger side (half the viewport minus the trigger, the most a
-block-centered trigger can guarantee), so at least one of the pair always
-fits — for every trigger position, including corners and a
-`--uktdd-body-min-height` above half the viewport, with no height cap
+can silently drop a fill. The two author fallbacks are the single-axis
```
acusti:
```suggestion
The default budget is two author fallbacks plus two named fill options,
`--uktdd-fill-bottom` and `--uktdd-fill-top`, so no engine
can silently drop a fill. The two author fallbacks are the single-axis
```

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -2,55 +2,51 @@
...
+flips of the primary (a block flip and an inline flip); the both-axes
+diagonal is dropped to stay within the floor. The cost of dropping the
```
acusti:
```suggestion
flips of the primary (a block flip and an inline flip); the opposite
diagonal flip is dropped to stay within the floor. The cost of dropping the
```

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -2,55 +2,51 @@
...
+diagonal is that a trigger crammed into the corner diagonal to the primary
+opening direction lands in a fill rather than a fourth natural placement,
+which for a corner-pinned trigger is a re-tune-the-primary case anyway.
```
acusti:
```suggestion
opening direction lands in a fill rather than a fourth natural placement.
```

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -2,55 +2,51 @@
...
+Each fill pairs a single-side `position-area` (`block-end`/`block-start`,
+spanning all inline tiles so it can never be rejected for horizontal
+overflow) with `justify-self: anchor-center`, which centers the body over
+the trigger and shifts it inward as needed to stay on-screen (the
+explicitness matters: `span-all`’s default alignment centers without that
+shift, so near a corner it would overflow and be rejected). A fill is only
+rejected when its side is shorter than the body’s min-block-size floor —
+what sends a cramped side’s fill to the roomier opposite side — and each
```
acusti:
```suggestion
rejected when its side is shorter than the body’s `min-block-size` floor, and each
```

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -2,55 +2,51 @@
...
+fill caps that floor at the worst-case larger side (roughly half the
```
acusti:
```suggestion
fill caps that floor at the larger side (roughly half the
```

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
```diff
@@ -2,55 +2,51 @@
...
+viewport, the most a block-centered trigger can guarantee), so at least one
+of the pair always fits — for every trigger position, including corners and
+a `--uktdd-body-min-height` above half the viewport, with no height cap
 needed. The pair is appended via the new `--uktdd-body-fill-fallbacks`
```
acusti:
```suggestion
of the pair always fits, no additional height cap needed.
The pair is appended via the new `--uktdd-body-fill-fallbacks`
```

.changeset/dropdown-split-fallback-slots.md
```diff
@@ -0,0 +1,30 @@
...
+
+The author fallbacks sit between the primary placement and the appended
+fills, and the component’s budget is exactly two of them (two author
+fallbacks + two fills = four past the base = the spec’s guaranteed
```
acusti:
```suggestion
fallbacks + two fills = four, plus the base = the spec’s guaranteed
```

.changeset/dropdown-split-fallback-slots.md
```diff
@@ -0,0 +1,30 @@
...
+five-option floor). As a comma-separated list, that limit was only
+documented, and a third entry silently pushed a fill past the floor —
```
acusti:
```suggestion
documented, and a third entry silently pushed a fill past the floor,
```

packages/dropdown/src/Dropdown.css
```diff
@@ -18,7 +18,14 @@
     --uktdd-body-pad-top: 0.5em;
     --uktdd-body-min-width: min(50px, 100%);
     --uktdd-body-position-area: block-end span-inline-end;
-    --uktdd-body-position-try-fallbacks: --uktdd-top-start, --uktdd-bottom-end;
+    /* the two author fallbacks, tried in order after the primary placement
```
acusti: for multi-sentence comments, we should lead with a capital letter 
```suggestion
    /* The two author fallbacks, tried in order after the primary placement
```

packages/dropdown/src/Dropdown.css
```diff
@@ -18,7 +18,14 @@
...
+       and before the fills. The defaults are the primary’s single-axis
+       flips: slot 1 the block flip, slot 2 the inline flip. Change direction
+       by resetting --uktdd-body-position-area and both slots together (see
+       the recipes in the README) — the primary sets the direction, the
```
acusti:
```suggestion
       the recipes in the README). The primary sets the direction, the
```

packages/dropdown/src/Dropdown.css
```diff
@@ -232,24 +240,22 @@
...
-           parent — is out of scope for the narrow columns of a macOS-style
-           menu; a consumer who needs it can append the block-cover fills to
-           --uktdd-submenu-fill-fallbacks */
+        /* like the body’s fill model, but a too-tall submenu stays beside
+           its parent — macOS never covers a parent menu with its submenu.
+           The list is the author fallback (the natural opposite placement),
+           then --uktdd-submenu-fill-fallbacks: full-height anchor-centered
+           fills, one per inline side, kept separate from
+           --uktdd-body-fill-fallbacks so a body’s upward flip doesn’t reorder
+           its submenus. These fills are rejected only on the inline axis, so
```
acusti: how would “a body’s upward flip reorder its submenus”? that’s a confusing concept to me, so if it isn’t important, let’s just cut it.

packages/dropdown/src/Dropdown.css
```diff
@@ -232,24 +240,22 @@
...
+           — unlike the body pair, whose two sides sum to the viewport —
+           there’s no at-least-one-always-fits guarantee: a submenu too tall
+           to open beside its parent and too wide for either side reverts to
+           its (overflowing) primary. Narrow macOS-style columns don’t hit
+           that; append the block-cover fills (--uktdd-fill-bottom,
+           --uktdd-fill-top) to --uktdd-submenu-fill-fallbacks if you need
+           it */
```
acusti:
```suggestion
           its (overflowing) primary. Normal submenus won’t hit that, but for
           extra wide menus, append the block-cover fills (--uktdd-fill-bottom,
           --uktdd-fill-top) to --uktdd-submenu-fill-fallbacks */
```

packages/dropdown/src/Dropdown.css
```diff
@@ -344,31 +359,25 @@ div.uktdropdown-body {
...
+   so a max-block-size clamp that would make it fit still reads as
+   overflowing and is never selected (same quirk that rejects max-content
+   here). 100% is the attached side’s space, less the symmetric
+   --uktdd-body-gap margins.
+
+   min-block-size still applies — a side shorter than --uktdd-body-min-height
+   is rejected and the other tried — but capped at the worst-case larger
+   side, (100dvb - anchor)/2 (a block-centered anchor splits the rest of the
+   viewport evenly), so a min above ~half the viewport can’t reject both
+   sides and strand the body at the overflowing primary.
```
acusti:
```suggestion
   min-block-size still applies (a side shorter than --uktdd-body-min-height
   is rejected and the other tried), but is capped at the worst-case larger
   side, so a min above ~half the viewport can’t reject both sides and
   strand the body at the overflowing primary.
```

packages/dropdown/src/Dropdown.css
```diff
@@ -344,31 +359,25 @@ div.uktdropdown-body {
...
+
+   The pair must be unrejectable on the inline axis so whichever block side
+   fits is always accepted: a bare block keyword spans all inline tiles
```
acusti:
```suggestion
   fits can be accepted. A bare block keyword spans all inline tiles
```

packages/dropdown/README.md
```diff
@@ -553,15 +555,20 @@ placement and sizing control when a product surface needs it.
...
+fits in the viewport — that’s what actually determines the direction a
+dropdown opens by default. The fallback slots are only consulted when the
+primary placement doesn’t fit, so overriding them alone changes nothing in
+the common case where the primary already fits. Overriding the primary
+alone works, but leaves behind fallback slots tuned for the old primary —
+in a cramped viewport, the dropdown can fall back toward the direction you
+just moved away from — so set all three together, and update both slots. If
+the new direction has only one meaningful fallback (a centered menu, say),
+put it in slot 1 and set slot 2 to the shipped no-op `--uktdd-noop` so it
+doesn’t inherit that flip.
```
```suggestion
dropdown opens by default. The fallback slots are tried in order only when the
primary placement doesn’t fit. Set all three together to ensure the menu behaves
the way you want to in all possible contexts. If the new direction has only one
meaningful fallback (a centered menu, say), put it in slot 1 and set slot 2 to the
shipped no-op `--uktdd-noop` so it doesn’t inherit that flip.
```

packages/dropdown/README.md
```diff
@@ -576,15 +583,18 @@ physical keyword on one axis with a logical one on the other — though
 `block-start`/`block-end` read as `top`/`bottom` in the near-universal
 horizontal-tb writing mode.
 
-Pick the row matching the direction you want, and set both variables to its
-pair:
+Pick the row matching the direction you want, and set the three variables
+to its values:
```
```suggestion
Pick the row matching the direction you want, and set the three CSS custom
properties to the corresponding values:
```

packages/dropdown/README.md
```diff
@@ -639,22 +649,22 @@ trigger can guarantee — so a min-height above half the viewport (honored in
...
+separate custom properties rather than a list: to avoid a third (or more)
+entry, which could result in a trigger near a viewport edge opening
+off-screen instead of flipping and scrolling. The two default slots are the
+single-axis flips of the primary (a block flip and an inline flip); the
+both-axes diagonal is intentionally not a slot, so a trigger crammed into
+the corner diagonal to the primary opening direction lands in a fill rather
+than a natural placement — re-tune the primary (per the recipes above) for
+triggers pinned to a viewport corner. For a recipe with a single meaningful
+fallback (a centered menu, say), set slot 1 and leave slot 2 as the shipped
+no-op `--uktdd-noop`, so it doesn’t inherit the default’s flip.
```
```suggestion
There are exactly two author fallback slots by design. The [CSS spec](https://drafts.csswg.org/css-anchor-position-1/#fallback-apply)
requires implementations to support at least five position options,
including the element’s base position, and `@acusti/dropdown` appends
two last-resort placements to the end, so that leaves two fallbacks that
you can specify, hence why the are exposed as two separate custom
properties. The two default slots are the
single-axis flips of the primary (a block flip and an inline flip); the
mirror opposite diagonal is intentionally not a slot, so that a trigger crammed into
the corner diagonal to the primary opening direction lands in a fill rather
than a natural placement. See the [recipes above](#changing-the-default-direction)
for different common placement scenarios. For a recipe with a single meaningful
fallback (a centered menu, say), set slot 1 and leave slot 2 as the shipped
no-op `--uktdd-noop`, so it doesn’t inherit the default’s flip.
```

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Applied in 38351e75.

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Applied in 38351e75.

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Applied in 38351e75.

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Applied in 38351e75.

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Applied in 38351e75.

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Applied in 38351e75.

.changeset/dropdown-fill-fallbacks-within-evaluation-limit.md
acusti: Applied in 38351e75.

.changeset/dropdown-split-fallback-slots.md
acusti: Applied in 38351e75.

.changeset/dropdown-split-fallback-slots.md
acusti: Applied in 38351e75.

packages/dropdown/src/Dropdown.css
acusti: Applied in 38351e75.

packages/dropdown/src/Dropdown.css
acusti: Applied in 38351e75.

packages/dropdown/src/Dropdown.css
acusti: Cut it in 38351e75 — you're right, it's a confusing tangent and the decoupling is an implementation detail that doesn't earn the reader's attention here.

packages/dropdown/src/Dropdown.css
acusti: Applied in 38351e75.

packages/dropdown/src/Dropdown.css
acusti: Applied in 38351e75 — also dropped the (100dvb - anchor)/2 formula, since your rewrite makes the point without it.

packages/dropdown/src/Dropdown.css
acusti: Applied in 38351e75.

packages/dropdown/README.md
acusti: Applied in 38351e75.

packages/dropdown/README.md
acusti: Applied in 38351e75.

packages/dropdown/README.md
acusti: Applied in 38351e75 (fixed one typo in the suggestion: 'hence why the are exposed' → 'they are exposed').

packages/dropdown/src/Dropdown.css
```diff
@@ -232,24 +240,20 @@
...
-           the full-height anchor-centered fills, one per inline side. Three
-           fallbacks past the base, within the spec’s five-option floor (the
-           block-cover pair is gone, so no engine reliance), and decoupled
-           from --uktdd-body-fill-fallbacks so a body’s upward flip no longer
-           reorders its submenus. The one case this drops versus covering the
-           parent — a submenu wider than the space on either side of its
-           parent — is out of scope for the narrow columns of a macOS-style
-           menu; a consumer who needs it can append the block-cover fills to
+        /* like the body’s fill model, but a too-tall submenu stays beside
+           its parent — macOS never covers a parent menu with its submenu.
```
acusti:
```suggestion
           its parent so that it can’t cover its parent menu.
```

.changeset/dropdown-split-fallback-slots.md
```diff
@@ -0,0 +1,30 @@
...
+previously set the two entries of `--uktdd-body-position-try-fallbacks`
+(the direction-recipe table in the README lists the per-direction values),
+and `--uktdd-submenu-position-try-fallback` (singular) where you set the
+submenu list.
+
+Also adds a shipped no-op placement, `--uktdd-noop`, for a recipe with a
+single meaningful fallback (a centered menu, say): set slot 1 and leave
+slot 2 as `--uktdd-noop` so it doesn’t inherit the default’s flip. It
+applies no overrides, so it evaluates as the primary placement — already
+rejected by the time it’s reached — and is skipped.
```
acusti: nit:
```suggestion
applies no overrides, so it evaluates as the primary placement and is skipped.
```

acusti: Applied in 96283213.

packages/dropdown/src/Dropdown.css
acusti: Applied in 96283213.

.changeset/dropdown-rename-position-try-fallbacks.md
```diff
@@ -40,11 +40,11 @@ names in `--uktdd-body-position-try-fallbacks`, or set
         --uktdd-top-right, --uktdd-bottom-left, --uktdd-top-left;
 }
```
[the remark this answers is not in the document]
acusti: Good catch — fixed in 80b45c51. This was an inconsistency I introduced when reconciling the sibling changesets: I updated the "after" to the slot form but left the original 3-entry "before", so it read as if the rename dropped a fallback. The "before" is now two entries mapping 1:1 to the two slots, keeping this changeset about the rename/logical-keyword migration alone.

PR #408, opened 2026-07-23
Make @acusti/dropdown default colors dark-mode friendly

### LINKED ISSUE #380: Make @acusti/dropdown dark mode friendly
update the default styles in @acusti/dropdown to work equally well in both a light mode and a dark mode color scheme (based on native browser primitives, no need to try to conform to non-standard dark/light mode implementations)

acusti:
## Summary

Closes #380. Updates `@acusti/dropdown`'s default styles to work equally well in light and dark mode, using native browser primitives rather than any bespoke light/dark implementation:

- Swaps hardcoded light-mode colors (`#fff`, `rgb(105, 162, 249)`, `rgba(0, 0, 0, …)`) for CSS system colors (`Canvas`, `CanvasText`, `Highlight`, `HighlightText`) and `color-mix()` blends against `CanvasText` for tints/shadows (`--uktdd-body-bg-color`, `--uktdd-body-bg-color-hover`, `--uktdd-body-color-hover`, `--uktdd-body-bg-color-selected`, `--uktdd-body-bg-color-path`, `--uktdd-menubar-trigger-bg-color-active`, and the body/submenu box-shadow).
- Adds a new `--uktdd-color-scheme` custom property (default `inherit`), applied via `color-scheme` on `.uktdropdown-body`, so the popover matches whatever `color-scheme` the host page itself has opted into — light, if the page hasn't adopted dark mode — rather than always following the OS/browser preference regardless of the page. Set it to `light dark` (or `light`/`dark`) to follow the user's preference irrespective of the host page instead.
- Fixes the body's `color: inherit` (which could pick up a fixed, non-scheme-aware text color from an ancestor) to read from `--uktdd-body-color` instead, matching the submenu's existing approach.
- Documents the new behavior in a "Color Scheme" README section and adds a patch changeset.

## Test plan

- [x] `bun run build`
- [x] `bun run test` (102 passing)
- [x] `bun run lint`
- [x] `bun run tsc`
- [x] `bun run format`
- [x] Visually verified in Storybook via Playwright, emulating both `light` and `dark` `colorScheme`, across a plain dropdown, a menubar, and nested submenus (including the muted "active path" highlight on an ancestor item)

---
_Generated by [Claude Code](https://claude.ai/code/session_01WBbX4XLzt9WnPdjVfpPrAT)_

<a href="https://cubic.dev/pr/acusti/uikit/pull/408?utm_source=github" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"&gt;``</a>

### DIFF
--- a/.changeset/dropdown-dark-mode-colors.md
+++ b/.changeset/dropdown-dark-mode-colors.md
@@ -1,0 +1,27 @@
+---
+'@acusti/dropdown': patch
+---
+
+Make the default colors work equally well in light and dark mode. The
+`--uktdd-body-bg-color`, `--uktdd-body-bg-color-hover`,
+`--uktdd-body-color-hover`, `--uktdd-body-bg-color-selected`, and
+`--uktdd-body-bg-color-path` now derive from CSS system colors (`Canvas`,
+`CanvasText`, `Highlight`, `HighlightText`) and `color-mix()` blends
+against `CanvasText`, instead of fixed light-mode values like `#fff`. The
+body’s and submenu’s previously untokenized `box-shadow` declarations are
+now exposed as `--uktdd-body-box-shadow`, kept as a fixed
+`rgba(0, 0, 0, 0.25)` rather than a scheme-aware mix — mixing in
+`CanvasText` turns it into a diffuse white glow in dark mode instead of a
+shadow, so a fixed black (subtle in dark mode, conventional in light mode)
+is the safer default in both. `--uktdd-menubar-trigger-bg-color-active`
+blends against `currentColor` instead, so its tint follows the trigger’s
+own authored foreground rather than the ambient color scheme.
+`color-scheme` (via the new `--uktdd-color-scheme` custom property, default
+`inherit`) is exposed so the popover can opt into `light dark` and follow
+the browser/OS preference regardless of the host page — it defaults to
+inheriting the host page’s own `color-scheme` instead, since the popover is
+page content rather than OS chrome. The body’s text color now reads from
+`--uktdd-body-color` instead of `color: inherit`, so it no longer picks up
+a fixed, non-scheme-aware text color from a shared ancestor. This changes
+the rendered colors of a dropdown using the defaults; override the custom
+properties to keep a previous look.
--- a/packages/docs/stories/Dropdown.css
+++ b/packages/docs/stories/Dropdown.css
@@ -5,45 +5,31 @@
 }
 
 /* Canvas/CanvasText so the demo canvas follows whatever color-scheme ends
-   up in effect — either the real ambient OS/browser preference (see the
-   media query below) or the CSSLengthsDropdown toggle. Native form
-   controls (inputs, buttons) already auto-dark from the OS preference
-   regardless of any page-level opt-in, so without this, a visitor with a
-   dark OS ends up with a native-dark input stranded on a hardcoded-light
-   page. :has(.uktdropdown) scopes this to Dropdown stories rather than
-   every Storybook page sharing these container id/class names */
-#storybook-root:has(.uktdropdown),
+   up in effect — either the real ambient OS/browser preference or the
+   ColorSchemeToggle. Native form controls (inputs, buttons)
+   already auto-dark from the OS preference regardless of any page-level
+   opt-in, so without this, a visitor with a dark OS ends up with a
+   native-dark input stranded on a hardcoded-light page. :has(.uktdropdown)
+   scopes this to Dropdown stories rather than every Storybook page
+   sharing these container id/class names */
 .docs-story:has(.uktdropdown) {
     background-color: Canvas;
     color: CanvasText;
 }
 
-/* a story with parameters.docs.story.inline: false (FillsWhenTooTall,
-   CoverFillRecipe, SubmenuFillsBesideParent) renders in its own isolated
-   iframe on the docs page rather than inline in the shared document —
-   #storybook-root there is a normal Canvas-view root, sized to its (short)
-   content by default, leaving the rest of the iframe’s allocated height
-   as unstyled white. Stretching it to the full viewport means the
-   Canvas background actually covers that iframe instead of leaving a gap
-   below the dropdown. Isolated stories have no CSSLengthsDropdown-toggle
-   awareness at all — they’re a separate document — so they always render
-   from the real ambient preference regardless of the toggle, same as this
-   rule’s scope implies */
-#storybook-root:has(.uktdropdown) {
-    min-height: 100dvh;
+/* the isolated iframe stories (inline: false) and the standalone Canvas
+   view are their own host page, with no color-scheme opt-in anywhere else
+   in the document — declare it directly on the actual document root
+   (rather than #storybook-root, which never matches on the docs page:
+   #storybook-root there is hidden and empty) so Canvas/CanvasText follow
+   the real OS/browser preference, and its background covers the full
+   viewport instead of just #storybook-root’s own (possibly short) box */
+:root:has(#storybook-root .uktdropdown) {
+    height: 100dvh;
+    color-scheme: light dark;
+    color: CanvasText;
+    background-color: Canvas;
 }
-
-/* No color-scheme override here for the rest of the (inline) Dropdown
-   stories: they deliberately inherit it from the document root, same as
-   any consumer page’s content would. CSSLengthsDropdown’s toggle sets it
-   there directly (see the story), so every other inline story on the docs
-   page visibly follows along too — the point of the demo is showing that
-   an unstyled dropdown picks up whatever color-scheme is in effect via
-   inheritance, and letting that inheritance actually reach the rest of
-   the page makes that concrete instead of demonstrating it in a single
-   isolated box. The toggle only runs when the visitor’s real OS/browser
-   prefers light (see the story), so this never fights the ambient
-   preference — it only ever moves everything the same direction, together */
 
 #storybook-root {
     &:has(.mk-header) {
--- a/packages/docs/stories/Dropdown.css
+++ b/packages/docs/stories/Dropdown.css
@@ -4,6 +4,47 @@
     flex-direction: column;
 }
 
+/* Canvas/CanvasText so the demo canvas follows whatever color-scheme ends
+   up in effect — either the real ambient OS/browser preference (see the
+   media query below) or the CSSLengthsDropdown toggle. Native form
+   controls (inputs, buttons) already auto-dark from the OS preference
+   regardless of any page-level opt-in, so without this, a visitor with a
+   dark OS ends up with a native-dark input stranded on a hardcoded-light
+   page. :has(.uktdropdown) scopes this to Dropdown stories rather than
+   every Storybook page sharing these container id/class names */
+#storybook-root:has(.uktdropdown),
+.docs-story:has(.uktdropdown) {
+    background-color: Canvas;
+    color: CanvasText;
+}
+
+/* a story with parameters.docs.story.inline: false (FillsWhenTooTall,
+   CoverFillRecipe, SubmenuFillsBesideParent) renders in its own isolated
+   iframe on the docs page rather than inline in the shared document —
+   #storybook-root there is a normal Canvas-view root, sized to its (short)
+   content by default, leaving the rest of the iframe’s allocated height
+   as unstyled white. Stretching it to the full viewport means the
+   Canvas background actually covers that iframe instead of leaving a gap
+   below the dropdown. Isolated stories have no CSSLengthsDropdown-toggle
+   awareness at all — they’re a separate document — so they always render
+   from the real ambient preference regardless of the toggle, same as this
+   rule’s scope implies */
+#storybook-root:has(.uktdropdown) {
+    min-height: 100dvh;
+}
+
+/* No color-scheme override here for the rest of the (inline) Dropdown
+   stories: they deliberately inherit it from the document root, same as
+   any consumer page’s content would. CSSLengthsDropdown’s toggle sets it
+   there directly (see the story), so every other inline story on the docs
+   page visibly follows along too — the point of the demo is showing that
+   an unstyled dropdown picks up whatever color-scheme is in effect via
+   inheritance, and letting that inheritance actually reach the rest of
+   the page makes that concrete instead of demonstrating it in a single
+   isolated box. The toggle only runs when the visitor’s real OS/browser
+   prefers light (see the story), so this never fights the ambient
+   preference — it only ever moves everything the same direction, together */
+
 #storybook-root {
     &:has(.mk-header) {
         height: 1000px;
@@ -41,6 +82,14 @@
 .uktdropdown input {
     box-sizing: border-box;
     width: 110px;
+}
+
+/* the UA default placeholder opacity reads as low-contrast in either
+   scheme; GrayText is the system color for de-emphasized text and already
+   tracks light/dark on its own */
+.uktdropdown-trigger::placeholder {
+    color: GrayText;
+    opacity: 1;
 }
 
 .uktdropdown-body {
@@ -88,7 +137,7 @@
             top: 10px;
             left: 4px;
             border-radius: 2px 0 0 3px;
-            background-color: #666;
+            background-color: currentColor;
             transform: rotateZ(38deg);
         }
 
@@ -101,7 +150,7 @@
             top: 10px;
             right: 4px;
             border-radius: 0 2px 2px 0;
-            background-color: #666;
+            background-color: currentColor;
             transform: rotateZ(-38deg);
         }
     }
--- a/packages/docs/stories/Dropdown.stories.tsx
+++ b/packages/docs/stories/Dropdown.stories.tsx
@@ -59,22 +59,90 @@ export default meta;
 type Story = StoryObj<typeof Dropdown>;
 
 export const CSSLengthsDropdown: Story = {
-    args: {
-        children: (
-            <ul>
-                <li>0px</li>
-                <li>4px</li>
-                <li>9px</li>
-                <li>18px</li>
-                <li>36px</li>
-                <li>54px</li>
-                <li>72px</li>
-                <li>144px</li>
-                <li>167px</li>
-                <li>198px</li>
-            </ul>
-        ),
-        className: 'css-lengths no-trigger-text',
+    parameters: {
+        docs: {
+            description: {
+                story: 'The dark/light toggle sets `color-scheme` on the document root, standing in for a host page that has (or hasn’t) opted into dark mode — the dropdown’s default colors follow it via inheritance rather than the OS/browser preference directly. Every other Dropdown story on this page inherits from that same document root, so toggling it moves them all together (stories that render in their own iframe are a separate document and always follow your real OS/browser preference instead). See [Color Scheme](#color-scheme) in the README. Disabled when your OS/browser already prefers dark mode, since from there the only direction left to demonstrate is toggling away from your actual preference, which wouldn’t show anything a real host page would do.',
+            },
+        },
+    },
+    render() {
+        const [colorScheme, setColorScheme] = React.useState<'dark' | 'light'>(() =>
+            window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light',
+        );
+        const [prefersDark, setPrefersDark] = React.useState(
+            () => window.matchMedia('(prefers-color-scheme: dark)').matches,
+        );
+
+        React.useEffect(() => {
+            const media = window.matchMedia('(prefers-color-scheme: dark)');
+            const onChange = () => {
+                // keep colorScheme in sync with ambient too, not just the
+                // disabled state: otherwise a mid-visit OS switch to dark
+                // disables the toggle but leaves a stale 'light' applied,
+                // which the disabled button can no longer correct
+                setPrefersDark(media.matches);
+                setColorScheme(media.matches ? 'dark' : 'light');
+            };
+            media.addEventListener('change', onChange);
+            return () => media.removeEventListener('change', onChange);
+        }, []);
+
+        React.useEffect(() => {
+            document.documentElement.style.colorScheme = colorScheme;
+            return () => {
+                document.documentElement.style.colorScheme = '';
+            };
+        }, [colorScheme]);
+
+        return (
+            <div style={{ position: 'relative' }}>
+                <button
+                    disabled={prefersDark}
+                    onClick={() =>
+                        setColorScheme(colorScheme === 'light' ? 'dark' : 'light')
+                    }
+                    style={{ position: 'absolute', right: 0, top: 0 }}
+                    title={
+                        prefersDark
+                            ? 'Switch your OS/browser to light mode to try this toggle'
+                            : undefined
+                    }
+                    type="button"
+                >
+                    {colorScheme === 'light'
+                        ? '🌙 Enable dark mode'
+                        : '☀️ Enable light mode'}
+                </button>
+                {prefersDark ? (
+                    <p
+                        style={{
+                            fontSize: '0.8em',
+                            margin: '0 0 0.5rem',
+                            paddingRight: '12em',
+                        }}
+                    >
+                        Your OS/browser already prefers dark mode, so this toggle is
+                        disabled — every other story on this page is already rendering
+                        from that same preference. Switch to light mode to try it live.
+                    </p>
+                ) : null}
+                <Dropdown className="css-lengths no-trigger-text">
+                    <ul>
+                        <li>0px</li>
+                        <li>4px</li>
+                        <li>9px</li>
+                        <li>18px</li>
+                        <li>36px</li>
+                        <li>54px</li>
+                        <li>72px</li>
+                        <li>144px</li>
+                        <li>167px</li>
+                        <li>198px</li>
+                    </ul>
+                </Dropdown>
+            </div>
+        );
     },
 };
 
--- a/packages/docs/stories/Dropdown.stories.tsx
+++ b/packages/docs/stories/Dropdown.stories.tsx
@@ -1,3 +1,11 @@
+import {
+    Controls,
+    Description,
+    Primary,
+    Stories,
+    Subtitle,
+    Title,
+} from '@storybook/addon-docs/blocks';
 import { fn } from 'storybook/test';
 import * as React from 'react';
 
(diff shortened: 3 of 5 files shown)


packages/dropdown/src/Dropdown.css
```diff
@@ -2,12 +2,29 @@
...
+       follow the user’s preference out of the box. Override to `light` or
+       `dark` to force a scheme irrespective of the user’s preference */
+    --uktdd-color-scheme: light dark;
+    /* Canvas/CanvasText/Highlight/HighlightText are UA system colors, so the
+       color defaults below need no prefers-color-scheme query of their own
+       to work in both schemes — they resolve against whichever scheme is
+       in effect for the element. A system color only switches to its dark
+       variant once a light-dark color-scheme is actually in effect though;
+       the UA default is light-only regardless of OS/browser preference
+       until something opts in, which is what --uktdd-color-scheme below is
```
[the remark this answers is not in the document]
acusti: Good catch, fixed in 037a08b (`below` → `above`).

---

packages/dropdown/src/Dropdown.css
```diff
@@ -58,7 +79,15 @@
...
-    --uktdd-menubar-trigger-bg-color-active: rgba(0, 0, 0, 0.12);
+    /* tints over whatever background the trigger already has, so a fixed
+       black wouldn’t register against a trigger the consumer’s already
+       styled dark; color-mix against CanvasText tints light instead of
+       dark when the ambient scheme is dark */
+    --uktdd-menubar-trigger-bg-color-active: color-mix(
+        in srgb,
+        CanvasText 12%,
+        transparent
+    );
```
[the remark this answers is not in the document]
acusti: Valid — fixed in 9f253ab by mixing against `currentColor` instead of `CanvasText`, so the tint follows the trigger's own authored foreground rather than the ambient color-scheme.

---

.changeset/dropdown-dark-mode-colors.md
```diff
@@ -0,0 +1,19 @@
...
+`CanvasText`, `Highlight`, `HighlightText`) and `color-mix()` blends
+against `CanvasText`, instead of fixed light-mode values like `#fff` and
+`rgba(0, 0, 0, …)`. The body also sets `color-scheme: light dark` (via the
+new `--uktdd-color-scheme` custom property) so it follows the browser/OS
+light-dark preference even on a host page that hasn't declared its own
+`color-scheme`, and its text color now reads from `--uktdd-body-color`
+instead of `color: inherit`, so it no longer picks up a fixed,
+non-scheme-aware text color from an ancestor. This changes the rendered
+colors of a dropdown using the defaults; override the custom properties to
+keep a previous look.
```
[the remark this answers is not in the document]
acusti: Valid — went with exposing a shadow token. Extracted `--uktdd-body-box-shadow` in 9f253ab so the body/submenu shadow is actually overridable via custom property, matching the changeset's migration note.

---

packages/dropdown/src/Dropdown.css
```diff
@@ -139,9 +168,19 @@
        top layer handles stacking (no z-index). The UA-popover box resets that
        must beat a consumer’s global [popover] styles live in the
        div.uktdropdown-body rule below; these cosmetics stay overridable */
-    color: inherit;
+    /* --uktdd-body-color, not inherit: the body still renders as a normal
+       DOM descendant of the trigger for inheritance purposes (top layer
+       only changes paint order), so a plain `inherit` would pick up
```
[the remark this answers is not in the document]
acusti: Correct — `.uktdropdown-body` and the trigger are both children of `.uktdropdown`, siblings not descendant. Fixed the comment in 9f253ab.

---

packages/dropdown/README.md
```diff
@@ -187,6 +187,24 @@ below. If your trigger sits near the right edge of the viewport, the
...
+`color-scheme: light dark` (via `--uktdd-color-scheme`) so it follows the
+browser/OS light-dark preference even when the host page hasn’t declared a
+`color-scheme` of its own — the dropdown looks native in both color schemes
+with no `prefers-color-scheme` query and no opt-in required. Override any
+of the `--uktdd-body-bg-color*` / `--uktdd-body-color*` /
+`--uktdd-body-box-shadow` custom properties to theme it instead, or set
+`--uktdd-color-scheme` to `light` or `dark` to pin the scheme regardless of
+the user’s preference. `--uktdd-menubar-trigger-bg-color-active` blends
+against `currentColor` instead, so it stays visible against a menubar
+trigger the consumer’s styled with its own background, independent of the
```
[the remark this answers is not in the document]
acusti: Fixed in 2b0260a.

---

packages/dropdown/src/Dropdown.css
```diff
@@ -58,7 +84,18 @@
...
-    --uktdd-menubar-trigger-bg-color-active: rgba(0, 0, 0, 0.12);
+    /* tints over whatever background the trigger already has, so a fixed
+       black wouldn’t register against a trigger the consumer’s already
+       styled dark; color-mix against currentColor (not CanvasText, which
+       tracks the ambient color-scheme rather than the trigger’s own
+       background) follows whatever foreground the trigger already uses for
+       contrast against its own background, so the tint stays visible
+       whether or not that matches the page’s color scheme */
+    --uktdd-menubar-trigger-bg-color-active: color-mix(
+        in srgb,
```
acusti: i like oklab/oklch

packages/dropdown/src/Dropdown.css
```diff
@@ -2,12 +2,34 @@
...
+       `dark` to force a scheme irrespective of the user’s preference */
+    --uktdd-color-scheme: light dark;
+    /* Canvas/CanvasText/Highlight/HighlightText are UA system colors, so the
+       color defaults below need no prefers-color-scheme query of their own
+       to work in both schemes — they resolve against whichever scheme is
+       in effect for the element. A system color only switches to its dark
+       variant once a light-dark color-scheme is actually in effect though;
+       the UA default is light-only regardless of OS/browser preference
+       until something opts in, which is what --uktdd-color-scheme above is
+       for */
```
acusti: i would simplify this to something like:
```suggestion
    /* Canvas(Text)/Highlight(Text) are UA system colors, so the color defaults
       below need no prefers-color-scheme query of their own to work in both
       schemes. The UA default is light-only regardless of OS/browser preference
       until something opts in, which is what --uktdd-color-scheme above is for */
```
(not sure about the line length, i guessed it)

packages/dropdown/src/Dropdown.css
```diff
@@ -2,12 +2,34 @@
...
-    --uktdd-body-bg-color: #fff;
-    --uktdd-body-bg-color-hover: rgb(105, 162, 249);
-    --uktdd-body-color-hover: #fff;
+    /* opts the body into the browser/OS light-dark preference (see the
+       color-scheme declaration on .uktdropdown-body) regardless of whether
+       the host page has opted in itself, so the system colors below — and
+       native form controls rendered inside the body, e.g. a checkbox item —
+       follow the user’s preference out of the box. Override to `light` or
+       `dark` to force a scheme irrespective of the user’s preference */
+    --uktdd-color-scheme: light dark;
```
acusti: this is cool, but i’m not sure if it’s a good idea. i could see an argument for it being more useful overall for the component to match the host page, rather than the OS. but an OS-level context-menu opened from the host page will render in the chosen OS color scheme regardless of the page’s color scheme, so maybe not?

after taking a beat to deliberate on it more, i’m thinking “match the host page” is the safer, and therefore more correct, default behavior. not sure what that means about using the `<system-color>` keywords. i suppose it still makes sense? are there any other clever things we should do to be a better host page citizen?

packages/dropdown/src/Dropdown.css
```diff
@@ -37,8 +59,12 @@
...
        submenus */
     --uktdd-body-gap: 0px;
-    --uktdd-body-color: canvastext;
-    --uktdd-body-bg-color-path: #dcdcdc;
+    --uktdd-body-color: CanvasText;
+    /* the muted highlight for an ancestor item on the active path (see the
+       [data-ukt-active] rules below); mixed against Canvas instead of a
+       fixed light gray so it reads as a gray step off the body background
+       in both a light and a dark Canvas */
+    --uktdd-body-bg-color-path: color-mix(in srgb, CanvasText 12%, Canvas);
```
acusti: i like oklab/oklch

packages/dropdown/src/Dropdown.css
```diff
@@ -2,12 +2,34 @@
...
-    --uktdd-body-bg-color-selected: rgba(0, 0, 0, 0.06);
+       selection), shown even after the active/hover highlight moves off it.
+       color-mix against CanvasText rather than a fixed black keeps the tint
+       readable on a dark Canvas too (light text mixed in, not dark) */
+    --uktdd-body-bg-color-selected: color-mix(in srgb, CanvasText 6%, transparent);
+    /* shared by the body and submenu (see their box-shadow declarations
+       below); color-mix against CanvasText rather than a fixed black keeps
+       it visible as a subtle glow rather than an imperceptible shadow
+       against a dark Canvas */
+    --uktdd-body-box-shadow: 0 8px 18px color-mix(in srgb, CanvasText 25%, transparent);
```
acusti: i like oklab/oklch

packages/dropdown/src/Dropdown.css
```diff
@@ -2,12 +2,34 @@
...
+    --uktdd-body-bg-color: Canvas;
+    --uktdd-body-bg-color-hover: Highlight;
+    --uktdd-body-color-hover: HighlightText;
     /* the persistent tint on the item matching props.value (the current
-       selection), shown even after the active/hover highlight moves off it */
-    --uktdd-body-bg-color-selected: rgba(0, 0, 0, 0.06);
+       selection), shown even after the active/hover highlight moves off it.
+       color-mix against CanvasText rather than a fixed black keeps the tint
+       readable on a dark Canvas too (light text mixed in, not dark) */
+    --uktdd-body-bg-color-selected: color-mix(in srgb, CanvasText 6%, transparent);
```
acusti: i like HDR colorspaces, especially oklab/oklch, which are now baseline widely available

packages/dropdown/src/Dropdown.css
```diff
@@ -2,12 +2,34 @@
...
+       for */
+    --uktdd-body-bg-color: Canvas;
+    --uktdd-body-bg-color-hover: Highlight;
+    --uktdd-body-color-hover: HighlightText;
     /* the persistent tint on the item matching props.value (the current
-       selection), shown even after the active/hover highlight moves off it */
-    --uktdd-body-bg-color-selected: rgba(0, 0, 0, 0.06);
+       selection), shown even after the active/hover highlight moves off it.
+       color-mix against CanvasText rather than a fixed black keeps the tint
+       readable on a dark Canvas too (light text mixed in, not dark) */
```
acusti: i think the color-mix is pretty self-explanatory
```suggestion
       selection), shown even after the active/hover highlight moves off it */
```

packages/dropdown/src/Dropdown.css
```diff
@@ -2,12 +2,34 @@
...
+    --uktdd-body-bg-color-selected: color-mix(in srgb, CanvasText 6%, transparent);
+    /* shared by the body and submenu (see their box-shadow declarations
+       below); color-mix against CanvasText rather than a fixed black keeps
+       it visible as a subtle glow rather than an imperceptible shadow
+       against a dark Canvas */
```
acusti: here also, the CanvasText-based color-mix is pretty self-explanatory
```suggestion
    /* shared by the body & submenu (see their box-shadow declarations below) */
```

packages/dropdown/src/Dropdown.css
```diff
@@ -139,9 +176,20 @@
...
        must beat a consumer’s global [popover] styles live in the
        div.uktdropdown-body rule below; these cosmetics stay overridable */
-    color: inherit;
+    /* --uktdd-body-color, not inherit: the body is the trigger’s DOM
+       sibling (both children of .uktdropdown), and top-layer paint order
+       doesn’t change inheritance, so a plain `inherit` would pick up
+       whatever fixed, non-scheme-aware text color their shared ancestors
+       happen to set — unreadable against a dark Canvas the moment that
+       color isn’t itself dark-mode aware. Matches the submenu’s explicit
+       color below */
```
acusti: now that we’ve introduced `--uktdd-body-color`, the fact that we use it as the `color` decl for the `.uktdropdown-body` selector doesn’t require explanation. also, the fact that the custom property is declared as `CanvasText` above is self-explanatory.

```suggestion
```

packages/dropdown/src/Dropdown.css
```diff
@@ -139,9 +176,20 @@
...
+    color: var(--uktdd-body-color);
+    /* inherited by [data-ukt-submenu] too, since it’s a DOM descendant of
+       the body (inheritance follows the DOM tree, not top-layer paint
+       order) */
+    color-scheme: var(--uktdd-color-scheme);
```
acusti, line 186: i’m guessing this is also implicated by the host page default preference i mentioned above

packages/dropdown/src/Dropdown.css
acusti: Agreed, and I think you talked yourself into the right answer — the OS context-menu analogy doesn't actually hold up, because that menu is browser/OS chrome, not page content. This popover, even rendered in the top layer, is still a normal descendant in the DOM tree for CSS purposes (top layer only changes paint order/containing-block, not inheritance), so "match the host page" is achievable for free: `color-scheme` is itself an inherited property.

Changed the default in 5faddaa: `--uktdd-color-scheme` is now `inherit` instead of `light dark`. Concretely:
- Host page hasn't declared `color-scheme` anywhere → popover stays light, even under a dark OS/browser preference (matches the page).
- Host page has `color-scheme: light dark` on its own root (which it basically needs anyway for its own native form controls/scrollbars to look right in dark mode) → the popover follows automatically via plain inheritance, no extra wiring needed.
- A consumer who wants the popover to follow the OS/browser preference regardless of what the page does can still set `--uktdd-color-scheme: light dark` explicitly — kept as a documented escape hatch.

Verified all three with Playwright against the `StatesDropdown` story (dark OS + no page opt-in → light; dark OS + page opts in → dark bg `rgb(18,18,18)`; explicit override → dark regardless of page).

On the system-color keywords: yes, still the right call — `Canvas`/`CanvasText`/`Highlight`/`HighlightText` just resolve against whichever `color-scheme` actually ends up in effect, so nothing about the color choices needed to change, only how the popover opts into a scheme.

Other host-page-citizenship ideas — I don't think there's more needed for this PR specifically. Everything's still override-able via custom properties (no `!important` anywhere), and `Highlight`/`HighlightText` already get you forced-colors-mode (Windows High Contrast) correctness for free since those are system colors too, so that's covered without extra work.

---

packages/dropdown/src/Dropdown.css
acusti: Switched to `oklab` in 5faddaa (across all four `color-mix()` calls).

---

packages/dropdown/src/Dropdown.css
acusti: Switched to `oklab` in 5faddaa.

---

packages/dropdown/src/Dropdown.css
acusti: Switched to `oklab` in 5faddaa. Went with oklab over oklch since every use here is tinting a single color against `transparent` (or `Canvas`) rather than mixing between two different hues, so the rectangular space sidesteps any hue-interpolation edge cases entirely.

---

packages/dropdown/src/Dropdown.css
acusti: Applied (with a small tweak since the default changed from `light dark` to `inherit`) in 5faddaa.

---

packages/dropdown/src/Dropdown.css
acusti: Applied in 5faddaa.

---

packages/dropdown/src/Dropdown.css
acusti: Applied in 5faddaa.

---

packages/dropdown/src/Dropdown.css
acusti: Removed in 5faddaa.

---

packages/dropdown/src/Dropdown.css
acusti: Yep — this line is `color-scheme: var(--uktdd-color-scheme)`, so it's the actual mechanism behind the default discussed above. Now inherits by default; see the reply on that thread for the full reasoning and verification.

---

packages/dropdown/src/Dropdown.css
acusti: nice, this sounds good to me. can we add a piece of UI to the default (top) Dropdown storybook story to have a dark/light mode switch in the top-right of the story that uses javascript to toggle the `color-scheme` on the root to serve as a demonstration of the new behavior?

.changeset/dropdown-dark-mode-colors.md
```diff
@@ -0,0 +1,24 @@
...
+`color-mix()` blends against `CanvasText`, instead of fixed light-mode
+values like `#fff` and `rgba(0, 0, 0, …)`.
+`--uktdd-menubar-trigger-bg-color-active` blends against `currentColor`
+instead, so its tint follows the trigger's own authored foreground rather
+than the ambient color scheme. `color-scheme` (via the new
+`--uktdd-color-scheme` custom property, default `inherit`) is exposed so
+the popover can opt into `light dark` and follow the browser/OS preference
+regardless of the host page — it defaults to inheriting the host page's own
+`color-scheme` instead, since the popover is page content rather than OS
+chrome. The body's text color now reads from `--uktdd-body-color` instead
```
acusti:
```suggestion
body and submenu’s previously untokenized `box-shadow`) now derive from CSS
system colors (`Canvas`, `CanvasText`, `Highlight`, `HighlightText`) and
`color-mix()` blends against `CanvasText`, instead of fixed light-mode
values like `#fff` and `rgba(0, 0, 0, …)`.
`--uktdd-menubar-trigger-bg-color-active` blends against `currentColor`
instead, so its tint follows the trigger’s own authored foreground rather
than the ambient color scheme. `color-scheme` (via the new
`--uktdd-color-scheme` custom property, default `inherit`) is exposed so
the popover can opt into `light dark` and follow the browser/OS preference
regardless of the host page — it defaults to inheriting the host page’s own
`color-scheme` instead, since the popover is page content rather than OS
chrome. The body’s text color now reads from `--uktdd-body-color` instead
```

packages/dropdown/src/Dropdown.css
acusti: Added in 4424194: `CSSLengthsDropdown` (the first story, top of the sidebar and the primary example on the autodocs page) now has a "🌙 Dark mode" / "☀️ Light mode" toggle button in the top-right, which sets `document.documentElement.style.colorScheme` on click and resets it on unmount. Verified with Playwright — clicking it flips the popover from light (white bg, dark text) to dark (`rgb(18,18,18)`-ish bg, light text) live, with the surrounding page area following along too since nothing else overrides `Canvas` on the story root.

---
_Generated by [Claude Code](https://claude.ai/code)

---

.changeset/dropdown-dark-mode-colors.md
acusti: Fixed in ab09b2b — straight apostrophes swapped for curly to match the rest of the docs.

---

packages/docs/stories/Dropdown.stories.tsx
```diff
@@ -59,22 +59,50 @@ export default meta;
...
+        return (
+            <div style={{ position: 'relative' }}>
+                <button
+                    onClick={() =>
+                        setColorScheme(colorScheme === 'light' ? 'dark' : 'light')
+                    }
+                    style={{ position: 'absolute', right: 0, top: 0 }}
+                    type="button"
+                >
+                    {colorScheme === 'light' ? '🌙 Dark mode' : '☀️ Light mode'}
```
acusti:
```suggestion
                    {colorScheme === 'light' ? '🌙 Enable dark mode' : '☀️ Enable light mode'}
```

acusti: Applied in 4bf16c6.

---

packages/docs/stories/Dropdown.css
```diff
@@ -4,6 +4,41 @@
...
 }
 
+/* Canvas/CanvasText so the demo canvas follows whatever color-scheme ends
+   up in effect — either the real ambient OS/browser preference (see the
+   media query below) or the CSSLengthsDropdown toggle. Native form
+   controls (inputs, buttons) already auto-dark from the OS preference
+   regardless of any page-level opt-in, so without this, a visitor with a
+   dark OS ends up with a native-dark input stranded on a hardcoded-light
+   page */
+#storybook-root,
```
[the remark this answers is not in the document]
acusti: Valid — fixed in f0c7a3f, scoping the base rule with `:has(.uktdropdown)` to match the media-query rule below it.

---

packages/docs/stories/Dropdown.stories.tsx
```diff
@@ -59,22 +59,87 @@ export default meta;
...
+        const [colorScheme, setColorScheme] = React.useState<'dark' | 'light'>(() =>
+            window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light',
+        );
+        const [prefersDark, setPrefersDark] = React.useState(
+            () => window.matchMedia('(prefers-color-scheme: dark)').matches,
+        );
+
+        React.useEffect(() => {
+            const media = window.matchMedia('(prefers-color-scheme: dark)');
+            const onChange = () => setPrefersDark(media.matches);
```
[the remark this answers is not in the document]
acusti: Valid — applied your suggested fix (syncing `colorScheme` alongside `prefersDark`) in ae12c16. The `.color-scheme-demo` exclusion this referenced is also gone as of the same commit — other inline Dropdown stories now deliberately inherit `color-scheme` from the document root instead of being pinned to ambient, so the toggle visibly moves them together rather than looking like it only affects one isolated example.

---
_Generated by [Claude Code](https://claude.ai/code)

---

packages/docs/stories/Dropdown.css
```diff
@@ -4,6 +4,47 @@
...
+   as unstyled white. Stretching it to the full viewport means the
+   Canvas background actually covers that iframe instead of leaving a gap
+   below the dropdown. Isolated stories have no CSSLengthsDropdown-toggle
+   awareness at all — they’re a separate document — so they always render
+   from the real ambient preference regardless of the toggle, same as this
+   rule’s scope implies */
+#storybook-root:has(.uktdropdown) {
+    min-height: 100dvh;
+}
+
```
acusti: i don’t mind the gap, let’s just drop this

packages/docs/stories/Dropdown.css
```diff
@@ -4,6 +4,47 @@
...
+   any consumer page’s content would. CSSLengthsDropdown’s toggle sets it
+   there directly (see the story), so every other inline story on the docs
+   page visibly follows along too — the point of the demo is showing that
+   an unstyled dropdown picks up whatever color-scheme is in effect via
+   inheritance, and letting that inheritance actually reach the rest of
+   the page makes that concrete instead of demonstrating it in a single
+   isolated box. The toggle only runs when the visitor’s real OS/browser
+   prefers light (see the story), so this never fights the ambient
+   preference — it only ever moves everything the same direction, together */
+
```
acusti: … and drop this

packages/docs/stories/Dropdown.css
acusti: Dropped in 79e8987, along with the adjacent "No color-scheme override" comment (your next thread). That cleanup also surfaced the actual bug behind it: the isolated `inline: false` stories weren't following the real OS dark preference at all — turns out `Canvas`/`CanvasText` only track `prefers-color-scheme` when something in the document explicitly opts in via `color-scheme: light dark`, and nothing did for those isolated iframes. Fixed by declaring it on `#storybook-root:has(.uktdropdown)`, which never touches the toggle-driven `.docs-story` boxes on the main docs page. Verified with Playwright that the isolated frame now actually renders `rgb(18, 18, 18)` under ambient dark (previously stayed white despite `color-scheme` computing correctly) and that the toggle's propagation to all 29 inline `.docs-story` boxes is unaffected.

---

packages/docs/stories/Dropdown.css
acusti: Dropped in 79e8987 — see the reply on the other thread for the fix that came out of this cleanup.

---

packages/docs/stories/Dropdown.css
```diff
@@ -4,6 +4,29 @@
...
+}
+
+/* #storybook-root here is the isolated iframe stories (inline: false) and
+   the standalone Canvas view, not the docs-page’s .docs-story boxes — they
+   have no host page opting into a color-scheme, so declare it directly to
+   get Canvas/CanvasText actually following the real OS/browser preference
+   instead of defaulting to light */
+#storybook-root:has(.uktdropdown) {
+    color-scheme: light dark;
+}
```
acusti: seems like maybe this should target the actual document `:root` instead of storybook root, and then we could also apply the appropriate background color:
```suggestion
:root:has(#storybook-root .uktdropdown) {
    color-scheme: light dark;
    height: 100dvh;
    color: canvastext;
    background-color: canvas;
}
```

acusti: Applied in 48b4e66, merged with the shared background/color declarations rather than duplicating them. Also folded in `height: 100dvh` on `:root` — that ends up giving us the isolated-iframe "gap" fix for free, since the root element's background paints the full viewport regardless of its own content height, so we no longer need a dedicated `min-height` rule on `#storybook-root` for that.

Verified `#storybook-root .uktdropdown` still only matches when a dropdown is actually present (confirmed the docs page's hidden `#storybook-root` is empty, so this rule doesn't leak onto the general docs page), and re-ran the isolated-frame/standalone-Canvas-view/toggle-propagation checks — all still pass.

---

packages/dropdown/src/Dropdown.css
```diff
@@ -2,12 +2,37 @@
...
-    --uktdd-body-bg-color-hover: rgb(105, 162, 249);
-    --uktdd-body-color-hover: #fff;
+    /* the popover is page content, not OS chrome — a native OS context menu
+       always renders in the OS’s own scheme regardless of the host page,
+       but this component should match whatever color-scheme the host page
+       itself declares (light, if the page hasn’t opted into dark mode),
+       so it inherits rather than forcing one. Override to `light dark` (or
+       `light`/`dark`) to follow the user’s OS/browser preference instead,
+       irrespective of the host page */
+    --uktdd-color-scheme: inherit;
```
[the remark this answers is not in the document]
acusti: Valid — the description was written before the default changed from `light dark` to `inherit` mid-review and never got updated. Fixed the PR description to match the code/README/changeset.

---

packages/docs/stories/Dropdown.stories.tsx
```diff
@@ -12,6 +20,78 @@ import type { Meta, StoryObj } from '@storybook/react-vite';
...
+        media.addEventListener('change', onChange);
+        return () => media.removeEventListener('change', onChange);
+    }, []);
+
+    React.useEffect(() => {
+        document.documentElement.style.colorScheme = colorScheme;
+        return () => {
+            document.documentElement.style.colorScheme = '';
+        };
+    }, [colorScheme]);
```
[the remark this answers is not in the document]
acusti: Checked — nothing else in the codebase ever sets `document.documentElement.style.colorScheme`, so there's no pre-existing inline value to preserve; capturing and restoring it would be functionally identical to resetting to `''` here. The reset-on-unmount is intentional (not a leftover default): it's specifically to avoid leaving the document in a toggled state if the visitor navigates away from this docs page to another story. Leaving as-is since there's no actual scenario in this app where the more defensive version would behave differently.

---

packages/dropdown/src/Dropdown.css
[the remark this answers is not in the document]
acusti: This is intentional, not a bug — see the earlier thread at discussion_r3639203314 for the full deliberation. The original design used `light dark` (follow the OS/browser regardless of host page), but we deliberately changed it to `inherit` after weighing it against the OS-context-menu analogy: a native context menu always follows the OS scheme because it's OS chrome, but this popover is page content in the DOM tree (top layer only changes paint order, not inheritance), so it should match whatever the host page itself has opted into — light, if the page hasn't adopted dark mode — same as any other element on that page would. A host page that wants the dropdown to follow the OS/browser preference regardless of its own scheme can still set `--uktdd-color-scheme: light dark` explicitly; that's the documented escape hatch, not the default. Leaving as-is.

---

.changeset/dropdown-dark-mode-colors.md
```diff
@@ -0,0 +1,27 @@
...
+'@acusti/dropdown': patch
+---
+
+Make the default colors work equally well in light and dark mode. The
+`--uktdd-body-bg-color`, `--uktdd-body-bg-color-hover`,
+`--uktdd-body-color-hover`, `--uktdd-body-bg-color-selected`, and
+`--uktdd-body-bg-color-path` now derive from CSS system colors (`Canvas`,
+`CanvasText`, `Highlight`, `HighlightText`) and `color-mix()` blends
+against `CanvasText`, instead of fixed light-mode values like `#fff`. The
+body and submenu’s previously untokenized `box-shadow` is now exposed as
```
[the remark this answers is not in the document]
acusti: Valid — fixed.

---

packages/docs/stories/Dropdown.css
```diff
@@ -4,6 +4,33 @@
     flex-direction: column;
 }
 
+/* Canvas/CanvasText so the demo canvas follows whatever color-scheme ends
+   up in effect — either the real ambient OS/browser preference or the
+   CSSLengthsDropdown toggle. Native form controls (inputs, buttons)
```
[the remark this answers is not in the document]
acusti: Valid — fixed.

---

packages/docs/stories/Dropdown.stories.tsx
```diff
@@ -12,6 +20,78 @@ import type { Meta, StoryObj } from '@storybook/react-vite';
...
+            setColorScheme(media.matches ? 'dark' : 'light');
+        };
+        media.addEventListener('change', onChange);
+        return () => media.removeEventListener('change', onChange);
+    }, []);
+
+    React.useEffect(() => {
+        document.documentElement.style.colorScheme = colorScheme;
+        return () => {
+            document.documentElement.style.colorScheme = '';
```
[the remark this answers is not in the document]
acusti: Same finding as Copilot's comment on this line (discussion_r3640828364) — see that reply: nothing else in the codebase ever sets `document.documentElement.style.colorScheme`, so there's no pre-existing inline value to preserve, and the reset-on-unmount is intentional. Leaving as-is for the same reason.

---

PR #411, opened 2026-08-04
Stop a disabled dropdown from opening via the keyboard

acusti:
**PR 1 of 6** in the pre-v1-RC a11y stack. Base: `main`.

## The bug

`props.disabled` was enforced only by `.uktdropdown.disabled`'s `pointer-events: none`, which stops the mouse and nothing else:

- the generated `<button>` trigger got no `disabled` attribute (only the searchable dropdown's `<input>` ever did, at `Dropdown.tsx:1305`)
- `handleKeyDown` never consulted the prop

So focusing a disabled dropdown's trigger and pressing Enter opened it. Verified before writing the fix:

```
BUTTON OUTER HTML: <button aria-controls="_r_0_" aria-expanded="false" aria-haspopup="menu"
                           class="uktdropdown-trigger" tabindex="0" type="button"></button>
disabled attr? false
aria-disabled? null
focused? true
opened after Enter? true
```

## The fix

- the generated button carries the native `disabled` attribute
- the three paths that *open* the dropdown — `handleKeyDown`, `handleMouseDown`, and `openOnHover`'s `handleDropdownMouseEnter` — check the prop

Gating the handlers rather than relying on the attribute alone is what covers a custom trigger, which can be any element and so may not support `disabled` at all. Such a trigger also picks up `aria-disabled`, filled in alongside the other trigger ARIA so a consumer-set value still wins.

Only *opening* is gated, so a dropdown disabled while already open still closes on Escape rather than trapping the user in an open menu.

## Tests

Six new cases under `describe('props.disabled')`. Five of them fail against the pre-fix source; the other two (Escape-still-closes, enabled-still-opens) are regression guards that pass either way.

`bun run test` (109 passed), `lint`, `tsc`, and `format` all clean.

---

<a href="https://cubic.dev/pr/acusti/uikit/pull/411?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/dropdown-disabled-blocks-opening.md
+++ b/.changeset/dropdown-disabled-blocks-opening.md
@@ -2,7 +2,7 @@
 '@acusti/dropdown': patch
 ---
 
-Stop a disabled dropdown from opening via the keyboard
+Stop a disabled dropdown from opening
 
 `props.disabled` was enforced only by `.uktdropdown.disabled`’s
 `pointer-events: none`, which stops the mouse and nothing else: the
@@ -10,9 +10,20 @@ generated trigger got no `disabled` attribute, and the key handler never
 consulted the prop, so focusing a disabled dropdown’s trigger and pressing
 Enter or Space opened it. The generated button now carries the native
 `disabled` attribute (as the searchable dropdown’s input already did), and
-the paths that open the dropdown — key, mousedown, and `props.openOnHover`
-hover — check the prop, so a custom trigger (which can be any element, and
-so can’t rely on the native attribute) is covered too. A custom trigger
+every path that opens the dropdown checks the prop, so a custom trigger —
+which can be any element, and so can’t rely on the native attribute — is
+covered too:
+
+- key, mousedown, and `props.openOnHover` hover
+- typing in a custom trigger’s text input, which `aria-disabled` doesn’t
+  make inert the way the native attribute makes the generated one
+- a `Menubar` sliding the open menu onto the member with ←/→, or switching
+  to it on hover, neither of which passes through the dropdown’s own
+  handlers
+
+A disabled `Menubar` member is now skipped rather than landed on, matching
+macOS, and ←/→ keep the current menu open when there’s no enabled member to
+move to instead of closing it and opening nothing. A custom trigger
 additionally receives `aria-disabled`, filled in like the other trigger
 ARIA, so a consumer-set value still wins. Only opening is gated: a dropdown
 disabled while already open still closes on Escape.
--- a/.changeset/dropdown-disabled-blocks-opening.md
+++ b/.changeset/dropdown-disabled-blocks-opening.md
@@ -4,26 +4,9 @@
 
 Stop a disabled dropdown from opening
 
-`props.disabled` was enforced only by `.uktdropdown.disabled`’s
-`pointer-events: none`, which stops the mouse and nothing else: the
-generated trigger got no `disabled` attribute, and the key handler never
-consulted the prop, so focusing a disabled dropdown’s trigger and pressing
-Enter or Space opened it. The generated button now carries the native
-`disabled` attribute (as the searchable dropdown’s input already did), and
-every path that opens the dropdown checks the prop, so a custom trigger —
-which can be any element, and so can’t rely on the native attribute — is
-covered too:
-
-- key, mousedown, and `props.openOnHover` hover
-- typing in a custom trigger’s text input, which `aria-disabled` doesn’t
-  make inert the way the native attribute makes the generated one
-- a `Menubar` sliding the open menu onto the member with ←/→, or switching
-  to it on hover, neither of which passes through the dropdown’s own
-  handlers
-
-A disabled `Menubar` member is now skipped rather than landed on, matching
-macOS, and ←/→ keep the current menu open when there’s no enabled member to
-move to instead of closing it and opening nothing. A custom trigger
-additionally receives `aria-disabled`, filled in like the other trigger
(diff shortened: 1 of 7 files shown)


.changeset/dropdown-disabled-blocks-opening.md
```diff
@@ -0,0 +1,29 @@
...
+- a `Menubar` sliding the open menu onto the member with ←/→, or switching
+  to it on hover, neither of which passes through the dropdown’s own
+  handlers
+
+A disabled `Menubar` member is now skipped rather than landed on, matching
+macOS, and ←/→ keep the current menu open when there’s no enabled member to
+move to instead of closing it and opening nothing. A custom trigger
+additionally receives `aria-disabled`, filled in like the other trigger
+ARIA, so a consumer-set value still wins. Only opening is gated: a dropdown
+disabled while already open still closes on Escape.
```
acusti: i appreciate the thoroughness but we don't need to go into all this detail. the fundamental thing is that disabled dropdowns won't open

packages/dropdown/README.md
```diff
@@ -267,6 +267,15 @@ type Props = {
...
     children: ReactNode | [ReactNode, ReactNode];
     className?: string;
+    /**
+     * Prevents the dropdown from opening by any means: pointer, Enter/Space,
+     * hover (props.openOnHover), typing in a custom trigger’s text input, and
+     * Menubar navigation onto it alike. The generated trigger carries the
+     * native disabled attribute; a custom trigger can be any element, so it
+     * receives aria-disabled instead. Inside a Menubar the member is skipped
+     * rather than landed on, the way macOS passes over a disabled menu. A
+     * dropdown disabled while already open still closes normally.
```
```suggestion
     * receives aria-disabled instead. Inside a Menubar the member is skipped.
     * A dropdown disabled while already open still closes normally.
```

packages/dropdown/src/Dropdown.tsx
```diff
@@ -90,6 +90,15 @@ export type Props = {
...
     children: ChildrenTuple | ReactElement;
     className?: string;
+    /**
+     * Prevents the dropdown from opening by any means: pointer, Enter/Space,
+     * hover (props.openOnHover), typing in a custom trigger’s text input, and
+     * Menubar navigation onto it alike. The generated trigger carries the
+     * native disabled attribute; a custom trigger can be any element, so it
+     * receives aria-disabled instead. Inside a Menubar the member is skipped
+     * rather than landed on, the way macOS passes over a disabled menu. A
+     * dropdown disabled while already open still closes normally.
```
```suggestion
     * receives aria-disabled instead. Inside a Menubar the member is skipped.
     * A dropdown disabled while already open still closes normally.
```

packages/dropdown/src/Dropdown.tsx
```diff
@@ -287,6 +296,7 @@ function RootDropdown({
 
     const allowCreateRef = useRef(allowCreate);
     const allowEmptyRef = useRef(allowEmpty);
+    const disabledRef = useRef(disabled);
```
acusti: do you think this earns its keep? our prop ref shaped pattern in this component was from pre-react compiler days

acusti: No, it doesn't — removed in 8896efd.

I'd added it out of caution rather than need, and your framing prompted me to actually check instead of reasoning about it. `handleInput` is created inside `handleRef`, which is an inline ref callback, so React tears it down and re-runs it on every render. The closure is rebuilt each time and can never be stale, which makes the ref exactly equivalent to reading the prop.

Rather than take that on faith I added a test that flips `disabled` on *after* mount — the only case where a stale closure could have differed — and it passes with the plain `disabled` read. That test stays in as the guard, so if the ref-callback behavior ever changes we find out.

On the broader point: you're right that the whole `*Ref` block reads as pre-compiler scaffolding. I didn't touch the rest of it here since it's outside what this PR is about, but a few of them look like they'd fall out the same way — the ones only read from handlers that are themselves recreated per render. Worth its own pass if you want it; happy to take that on separately rather than smuggle it into an a11y fix.

---

acusti (commented):
it would be good to have a test for the behavior where you type in the first letters of a menu item and it focuses the best matching item that shows that it works when the member is enabled but focuses a different item when the member is disabled

acusti:
Added in 6145541 — a pair, so the behavior reads by contrast rather than by assertion:

| items | typed | active |
|---|---|---|
| all enabled | `bu` | **Bullets** |
| `Bullets` disabled | `bu` | **Bold** |

The same keystrokes that reach Bullets reach Bold once Bullets is disabled. Typeahead builds its candidate list from `getItemElements`, which filters `[aria-disabled="true"]`, so a disabled item is never a match target and the best *remaining* candidate wins — rather than typeahead landing the highlight on something unselectable.

I checked the actual behavior before writing the assertions rather than predicting it, since "falls to the next best match" and "matches nothing" were both plausible and only one is true.

Two notes on what this is and isn't:

It characterizes behavior that predates this branch — the `getItemElements` filter was already there — so it's coverage, not a fix. It belongs here because it's part of the disabled contract this PR is about, and because the existing disabled-item coverage only exercised arrow navigation.

It's about disabled **items**, not `props.disabled` on the dropdown itself. If you meant a disabled *menubar member* — typeahead within a member's open menu — that one can't arise: a disabled member never opens, so there's no menu to type into. Say the word if you were after something different.

---

PR #413, opened 2026-08-04
Give the popup body an accessible name

acusti:
**PR 3 of 6** in the pre-v1-RC a11y stack. Base: #412 — review that first; this diff is only the third commit.

## The gap

The open body carried a `role` but no name. For `hasItems={false}` that means a `role="dialog"` with nothing naming it, which is an ARIA violation (axe `aria-dialog-name`); menus and listboxes were announced as a bare "menu"/"listbox". Verified before the fix:

```
DIALOG: <div class="uktdropdown-body" id="_r_5_" popover="manual" role="dialog">…
aria-label? null   aria-labelledby? null
```

## The fix

The body takes `aria-labelledby` pointing at whatever names the trigger:

1. the `props.label` text when there is one — it wins because it's always text and is already the name the trigger itself takes, so both announce the same thing
2. otherwise the trigger element, whose accessible name the body inherits

So a menu is announced as "Alignment menu" rather than "menu".

## The id scheme

Pointing at the trigger means the trigger needs an id, so this PR introduces the scheme the rest of the stack builds on: everything hangs off the one `useId()` the component already calls for the body, with a suffix naming its role (`-trigger`, `-label`). `useId` is unique per instance, so the suffixed ids collide with nothing and need no module-level counter. PR 4 extends the same scheme to item ids with an index path.

A custom trigger carrying its own `id` keeps it, and the body points at that one instead.

## One deliberate non-fix

A searchable dropdown with no `props.label` is left **unnamed** rather than named wrongly. `aria-labelledby` resolves a text input to its *value*, so pointing the listbox at the generated input would name it after whatever the user has typed. That dropdown has no accessible name for its trigger either, so `props.label` (or a custom trigger with its own `aria-label`) is the fix for both. There's a test pinning this so it isn't "fixed" into the value-naming behavior later.

## Tests

Six new cases under `describe('popup accessible name')`, asserting through `getByRole(…, { name })` so they exercise real accname computation rather than attribute equality. Five fail against the pre-fix source; the label-less-searchable one is the guard described above.

`bun run test` (119 passed), `lint`, `tsc`, and `format` all clean.

---

<a href="https://cubic.dev/pr/acusti/uikit/pull/413?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/dropdown-popup-accessible-name.md
+++ b/.changeset/dropdown-popup-accessible-name.md
@@ -1,0 +1,25 @@
+---
+'@acusti/dropdown': minor
+---
+
+Give the popup body an accessible name
+
+The open body carried a `role` but no name, so a `hasItems={false}`
+dropdown rendered a `role="dialog"` with nothing naming it — a dialog
+without an accessible name is an ARIA violation, and menus and listboxes
+were announced as a bare “menu”/“listbox” too. The body now takes
+`aria-labelledby` pointing at whatever names the trigger: the `props.label`
+text when there is one (always text, and the same name the trigger itself
+takes), otherwise the trigger element, whose accessible name the body then
+inherits.
+
+To point at the trigger the component fills in its `id`, derived from the
+same `useId` that already produced the body’s. A custom trigger that
+carries its own `id` keeps it, and the body points at that one instead.
+
+A searchable dropdown with no `props.label` is deliberately left unnamed:
+`aria-labelledby` resolves a text input to its value, so pointing the
+listbox at the generated input would name it after whatever the user has
+typed. Such a dropdown has no accessible name for its trigger either, so
+`props.label` — or a custom trigger carrying its own `aria-label` — is the
+fix for both.
--- a/.changeset/dropdown-popup-accessible-name.md
+++ b/.changeset/dropdown-popup-accessible-name.md
@@ -17,9 +17,14 @@ To point at the trigger the component fills in its `id`, derived from the
 same `useId` that already produced the body’s. A custom trigger that
 carries its own `id` keeps it, and the body points at that one instead.
 
-A searchable dropdown with no `props.label` is deliberately left unnamed:
-`aria-labelledby` resolves a text input to its value, so pointing the
-listbox at the generated input would name it after whatever the user has
-typed. Such a dropdown has no accessible name for its trigger either, so
-`props.label` — or a custom trigger carrying its own `aria-label` — is the
-fix for both.
+Two shapes are deliberately left unnamed, both because the trigger has no
+name to give. A text input trigger is one: `aria-labelledby` resolves a
+text input to its value, so pointing the popup at one would name it after
+whatever the user has typed. That covers a searchable dropdown’s trigger (a
+text input by definition, generated or custom) and a custom trigger that is
+itself an `<input>` or `<textarea>`; a text input nested inside a custom
+trigger leaks its value the same way but isn’t detectable from the element
+alone. The other is an empty generated trigger — single-child syntax with
+no `props.label` renders an empty `<button>`, which has no name of its own
+to lend, and for `hasItems={false}` that leaves an unnamed dialog. In both
+cases the trigger is unnamed too, and `props.label` fixes both at once.
--- a/.changeset/dropdown-popup-accessible-name.md
+++ b/.changeset/dropdown-popup-accessible-name.md
@@ -4,27 +4,12 @@
 
 Give the popup body an accessible name
 
-The open body carried a `role` but no name, so a `hasItems={false}`
-dropdown rendered a `role="dialog"` with nothing naming it — a dialog
-without an accessible name is an ARIA violation, and menus and listboxes
-were announced as a bare “menu”/“listbox” too. The body now takes
-`aria-labelledby` pointing at whatever names the trigger: the `props.label`
-text when there is one (always text, and the same name the trigger itself
-takes), otherwise the trigger element, whose accessible name the body then
-inherits.
+The open body now takes `aria-labelledby` pointing at whatever names the
+trigger — `props.label` if you pass one, otherwise the trigger element.
+Previously it carried a `role` but no name, so `hasItems={false}` rendered
+an unnamed `dialog`.
 
-To point at the trigger the component fills in its `id`, derived from the
-same `useId` that already produced the body’s. A custom trigger that
-carries its own `id` keeps it, and the body points at that one instead.
-
-Two shapes are deliberately left unnamed, both because the trigger has no
-name to give. A text input trigger is one: `aria-labelledby` resolves a
-text input to its value, so pointing the popup at one would name it after
-whatever the user has typed. That covers a searchable dropdown’s trigger (a
-text input by definition, generated or custom) and a custom trigger that is
-itself an `<input>` or `<textarea>`; a text input nested inside a custom
-trigger leaks its value the same way but isn’t detectable from the element
-alone. The other is an empty generated trigger — single-child syntax with
-no `props.label` renders an empty `<button>`, which has no name of its own
-to lend, and for `hasItems={false}` that leaves an unnamed dialog. In both
-cases the trigger is unnamed too, and `props.label` fixes both at once.
+Two shapes stay unnamed, because the trigger has no name to lend: a text
+input trigger (`aria-labelledby` resolves one to its value) and an empty
+generated trigger (single-child syntax with no `props.label`). Pass
+`props.label` to name either.
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -176,9 +176,7 @@ const CLICKABLE_SELECTOR = 'button, a[href], input[type="button"], input[type="s
 const FOCUSABLE_SELECTOR = 'a[href], button, input, select, textarea, [tabindex]';
 // Any input that isn’t one of the non-text types, plus textarea. Derived from
 // the same list that drives use-keyboard-events’ isEventTargetUsingKeyEvent, so
-// the two can’t disagree about what counts as a text input — the hand-written
-// version of this listed only radio, checkbox, and range, which meant a submit
-// or button input in a custom trigger was adopted as its value source.
+// the two can’t disagree about what counts as a text input.
 const TEXT_INPUT_SELECTOR = `input${NON_TEXT_INPUT_TYPES.map(
     (type) => `:not([type=${type}])`,
 ).join('')},textarea`;
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -1368,11 +1368,9 @@ function RootDropdown({
     //
     // A text input can’t serve as that source: aria-labelledby resolves one to
     // its *value*, so pointing at it would name the popup after whatever the
-    // user has typed. That rules out the trigger of a searchable dropdown —
-    // a text input by definition, generated or custom — and a custom trigger
-    // that is itself an input or textarea. Such a dropdown has no accessible
-    // name for its trigger either, so the fix for both is the same: pass
-    // props.label (or, for the trigger alone, a custom trigger’s aria-label).
+    // user has typed. That rules out triggers that are an input or textarea,
+    // which have no accessible name. The fix is to pass props.label (or, for
+    // the trigger alone, a custom trigger’s aria-label).
     let bodyLabelledBy: string | undefined;
 
     if (!isValidElement(trigger)) {
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -191,6 +191,33 @@ const SAFE_AREA_TIMEOUT = 300;
 // somewhere the pointer briefly leaves) doesn’t flicker-close it
 const HOVER_CLOSE_DELAY = 150;
 
+// Input types that hold no user-entered text, so aria-labelledby resolves them
+// by their label/content rather than their value (mirrors TEXT_INPUT_SELECTOR)
+const NON_TEXT_INPUT_TYPES = new Set([
+    'button',
+    'checkbox',
+    'color',
+    'file',
+    'hidden',
+    'image',
+    'radio',
+    'range',
+    'reset',
+    'submit',
+]);
+
+// Whether a custom trigger element is itself a text input — one aria-labelledby
+// would resolve to its value rather than to a name (see bodyLabelledBy). Only
+// the element itself is checked: a text input nested inside a custom trigger
+// leaks its value the same way, but that isn’t knowable from the element alone,
+// and props.label is the documented answer for both.
+const isTextInputElement = (element: ReactElement) => {
+    if (element.type === 'textarea') return true;
+    if (element.type !== 'input') return false;
+    const { type } = element.props as { type?: string };
+    return type == null || !NON_TEXT_INPUT_TYPES.has(type);
+};
+
 export default function Dropdown(props: Props) {
     const parentDropdown = useContext(DropdownContext);
     // A Dropdown nested inside a menu dropdown’s body renders as a submenu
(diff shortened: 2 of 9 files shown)


packages/dropdown/src/Dropdown.tsx
```diff
@@ -1362,18 +1382,26 @@ function RootDropdown({
         // element, so props.disabled is conveyed as aria-disabled rather than
         // the native attribute — the open paths enforce it either way.
         const triggerProps = trigger.props as Record<string, unknown>;
+        // Point the body at whatever id the trigger ends up with, so a
+        // consumer-set one is honored rather than overwritten
+        const resolvedTriggerId = (triggerProps.id as string | undefined) ?? triggerId;
+        bodyLabelledBy = resolvedTriggerId;
```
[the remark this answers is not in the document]
acusti: Valid, and the sharpest of the four — this was a regression I introduced, not just an unfixed gap: that shape previously had no `aria-labelledby` at all, and the custom-trigger branch gave it one pointing at a live input value.

Fixed in 66cc11d by moving the rule to where the reason lives instead of leaving it attached to one branch. A text input can't name the popup, so the body points at no trigger that is one:

- `isSearchable` — its trigger is a text input by definition, generated *or* custom, so the whole mode is covered rather than just the generated branch
- a custom trigger that is itself an `<input>` (excluding the non-text `type`s) or `<textarea>`, via a new `isTextInputElement` check

On "the component can't tell a custom trigger is a text input" — it can for the element itself, which is what the check does. What it can't see is a text input *nested* inside a custom trigger, which leaks its value identically. That one is documented rather than guarded, in both the README and the comment on `isTextInputElement`, pointing at `props.label`.

Four tests added: the searchable and non-searchable custom-input cases (both fail against the previous commit), plus guards that `props.label` still names a listbox behind a custom text input and that a non-text input trigger like `<input type="button">` still names the body normally.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1342,13 +1360,15 @@ function RootDropdown({
                 />
             );
         } else {
+            bodyLabelledBy = triggerId;
```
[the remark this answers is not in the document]
acusti: Valid, and worth stating in the docs — done in 66cc11d, but documented rather than fixed in code, for two reasons.

It isn't specific to `hasItems={false}`, and it isn't new. Single-child syntax with no `props.label` renders `<button>` with nothing inside it, so the **trigger** has no accessible name either — that's true today on `main`, for every popup role, and is arguably the more serious half. This PR doesn't regress it; it just can't fix it, because there's no name anywhere to derive one from. Deriving from the body content would name a dialog after its own contents, which is worse than unnamed.

Rejecting it isn't right either: the component already `console.error`s for genuine misuse (children count, inert submenu props), and this isn't misuse of the API so much as an incomplete call — and warning on every label-less single-child dropdown would fire on plenty of correct code where the consumer names the trigger some other way.

So the README's accessible-name section now lists both unnamed shapes together, with the point that the trigger is unnamed in each and `props.label` fixes both at once, and calls out that this leaves `hasItems={false}` with an unnamed dialog specifically.

Happy to add a dev-only warning for the `hasItems={false}` + no-label + single-child intersection if you'd rather it be loud — say the word and I'll fold it in.

---

.changeset/dropdown-popup-accessible-name.md
```diff
@@ -0,0 +1,30 @@
...
+name to give. A text input trigger is one: `aria-labelledby` resolves a
+text input to its value, so pointing the popup at one would name it after
+whatever the user has typed. That covers a searchable dropdown’s trigger (a
+text input by definition, generated or custom) and a custom trigger that is
+itself an `<input>` or `<textarea>`; a text input nested inside a custom
+trigger leaks its value the same way but isn’t detectable from the element
+alone. The other is an empty generated trigger — single-child syntax with
+no `props.label` renders an empty `<button>`, which has no name of its own
+to lend, and for `hasItems={false}` that leaves an unnamed dialog. In both
+cases the trigger is unnamed too, and `props.label` fixes both at once.
```
acusti: so many words, can we shrink this down? these should be skimmable while providing enough detail to ensure users don’t miss important points when considering an upgrade or evaluating a version

acusti: Cut from 30 lines to 15. Since this is the second time you've flagged changeset length (the `disabled` one earlier), I took it as a standing preference and applied it to the other three too rather than have you ask twice more:

| changeset | before | after |
|---|--:|--:|
| popup accessible name (this PR) | 30 | 15 |
| `aria-activedescendant` (#414) | 42 | 15 |
| combobox role (#415) | 29 | 17 |
| menubar semantics (#416) | 39 | 20 |

The rule I used, from your framing: keep what changed and anything that affects an upgrade decision; drop the reasoning and implementation narrative, which belongs in code comments and the README.

So what went was things like why `aria-autocomplete` is `"list"` rather than `"both"`, how the roving tab stop is claimed and reassigned, which call sites sync `aria-activedescendant`, and the ARIA 1.1 wrapper rationale. What stayed is the behavior change, plus the parts someone upgrading would be annoyed to discover on their own — the two shapes that stay unnamed here, generated item ids being internal and reassigned in #414, and the exposed-role changes in #415 and #416.

For those last two I put the role change under a bolded **Upgrading:** line, since "your `getByRole('textbox')` queries will break" is exactly the sort of thing that shouldn't need reading a paragraph to find. Happy to drop that convention if it reads as noise.

Pushed to each PR's own branch: 66d1e5a here, 8dde500 on #414, f28e297 on #415, 2ecd101 on #416.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -191,6 +191,33 @@ const SAFE_AREA_TIMEOUT = 300;
 // somewhere the pointer briefly leaves) doesn’t flicker-close it
 const HOVER_CLOSE_DELAY = 150;
 
+// Input types that hold no user-entered text, so aria-labelledby resolves them
+// by their label/content rather than their value (mirrors TEXT_INPUT_SELECTOR)
+const NON_TEXT_INPUT_TYPES = new Set([
```
acusti: this is more complete than `TEXT_INPUT_SELECTOR`. let’s move it to before `TEXT_INPUT_SELECTOR` and use it to construct the `:not(…)` chain so they are in agreement and we get the benefits of the better selector where needed.

wait, actually this is byte-for-byte equivalent to
```
const NON_TEXT_INPUT_TYPES = new Set([
    'button',
    'checkbox',
    'color',
    'file',
    'hidden',
    'image',
    'radio',
    'range',
    'reset',
    'submit',
]);
```
from `packages/use-keyboard-events/src/handlers.ts`, which this package already depends on, and which drives `isEventTargetUsingKeyEvent`. let’s update use-keyboard-events to export it and we can import and use it here in this file.

acusti: Confirmed byte-for-byte — same ten types, same order, `handlers.ts:72`. Done in a580e29: exported from `use-keyboard-events` (from `handlers.ts` and re-exported off the entry) and imported here, so there's one definition instead of two copies to keep in sync.

`TEXT_INPUT_SELECTOR` is now derived from it:

```js
const TEXT_INPUT_SELECTOR = `input${Array.from(
    NON_TEXT_INPUT_TYPES,
    (type) => `:not([type=${type}])`,
).join('')},textarea`;
```

Your first instinct was right about it being the better selector, and it turns out that wasn't only tidiness — it's a bug fix. The hand-written chain excluded just `radio`, `checkbox`, and `range`, so a `submit`, `button`, `file`, or `hidden` input inside a custom trigger was adopted as the dropdown's **value source** at `Dropdown.tsx:1200`. Added a test for it: `allowCreate` with an `<input type="submit">` trigger must still be a no-op on submit, since there's no text to create from. It fails against the old selector.

Two changesets rather than one, since the change spans packages: `minor` on `use-keyboard-events` for the new export, `patch` on `dropdown` for the widened selector. `updateInternalDependencies: "patch"` handles the `^0.11.1` range at release, so the cross-package import doesn't need a manual bump.

---

packages/use-keyboard-events/src/handlers.ts
```diff
@@ -69,7 +69,10 @@ export function isPrimaryModifierPressed(event: KeyboardEvent) {
     return IS_APPLE_REGEXP.test(platform) ? event.metaKey : event.ctrlKey;
 }
 
-const NON_TEXT_INPUT_TYPES = new Set([
+// The input types that hold no user-entered text. Exported because consumers
+// need the same notion of "is this a text input" that isEventTargetUsingKeyEvent
+// uses — keeping a second copy in sync by hand is how they drift apart.
+export const NON_TEXT_INPUT_TYPES = new Set([
```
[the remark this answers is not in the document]
acusti: Valid — it's a singleton every consumer shares, and nothing stopped a consumer from editing it out from under `isEventTargetUsingKeyEvent`. Fixed in b2d1013.

Of the two remedies suggested, only one of them works. `Object.freeze` does nothing to a `Set` — the internal slots aren't properties, so `.add()` still mutates a frozen one:

```js
const s = Object.freeze(new Set(['a']));
s.add('b'); // no throw
s.size;     // 2
```

`ReadonlySet<string>` would have stopped TypeScript consumers, but it's a type-level assertion with no runtime effect, and this package ships to JS consumers too. So the export is now a frozen array instead, which actually can't be mutated:

```ts
export const NON_TEXT_INPUT_TYPES = Object.freeze([...]);   // readonly string[]

// Membership is checked on every key event, so pay for the Set once here
const NON_TEXT_INPUT_TYPE_SET = new Set(NON_TEXT_INPUT_TYPES);
```

The private `Set` keeps `isEventTargetUsingKeyEvent`'s hot-path lookup O(1), and derives from the array rather than being a second hand-maintained copy, so the drift this export existed to prevent isn't reintroduced. The two consumer-side call sites move to `.map()` and `.includes()`; the generated `TEXT_INPUT_SELECTOR` string is byte-identical, since `Array.from` over a `Set` already walked insertion order.

Added a test asserting the export is frozen and that pushing to it throws, which fails if the `Object.freeze` is ever dropped. 145 tests pass across both packages; lint, tsc, and format clean.

---

packages/use-keyboard-events/src/handlers.ts
```diff
@@ -69,7 +69,13 @@ export function isPrimaryModifierPressed(event: KeyboardEvent) {
     return IS_APPLE_REGEXP.test(platform) ? event.metaKey : event.ctrlKey;
 }
 
-const NON_TEXT_INPUT_TYPES = new Set([
+// The input types that hold no user-entered text. Exported because consumers
+// need the same notion of "is this a text input" that isEventTargetUsingKeyEvent
+// uses — keeping a second copy in sync by hand is how they drift apart. Frozen,
+// and an array rather than a Set, because it's a singleton every consumer shares:
+// Object.freeze can't protect a Set (.add still mutates a frozen one), so an
+// array is the only shape that can't be edited out from under its readers.
```
acusti: all prose, including comments, should use typographic quotes (applies to all comments added in this PR). also, we don’t need to explain the choice of an array over a Set.
```suggestion
// the input types that cannot hold user-entered text, exported so consumers can share
// the notion of “is this a text input” (e.g. @acusti/dropdown)
```
lastly, i switched it to lowercase because house style is only to use sentence-case for multi-sentence comments, and this change makes it one sentence (also applies to all comments added in this PR, e.g. `// membership is checked on every key event, so pay for the Set once here` below)

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1337,6 +1361,20 @@ function RootDropdown({
...
+    // the best source when it’s there — it’s always text, and it’s the same
+    // name the trigger itself takes — so it wins; otherwise the body points at
+    // the trigger element and inherits its accessible name.
+    //
+    // A text input can’t serve as that source: aria-labelledby resolves one to
+    // its *value*, so pointing at it would name the popup after whatever the
+    // user has typed. That rules out the trigger of a searchable dropdown —
+    // a text input by definition, generated or custom — and a custom trigger
+    // that is itself an input or textarea. Such a dropdown has no accessible
+    // name for its trigger either, so the fix for both is the same: pass
```
```suggestion
    // user has typed. That rules out triggers that are an input or textarea, which have 
    // no accessible name. The is to pass
```
needs to be re-wrapped

packages/use-keyboard-events/src/handlers.ts
acusti: Applied in 3f0d4df, and swept the rest of the prose this PR adds rather than just this comment.

Your suggestion is in verbatim. The other three rules-driven changes here:

- **Typographic quotes** — the added comments were inconsistent, some already curly and some straight. Now uniform, including apostrophes in contractions (`isn't` → `isn’t`, `use-keyboard-events'` → `use-keyboard-events’`). Also covers the changeset prose, since the rule is about all prose.
- **Lowercase for single-sentence comments** — four of them: `// membership is checked on every key event…` as you called out, plus `// point the body at whatever id the trigger ends up with…` here and `// a submit input is not a text input…` in the test. The multi-sentence ones keep sentence-case. I also dropped a trailing period while lowercasing one, since the single-sentence comments in this file don't carry one — say the word if you'd rather keep it.
- **Dropped the array-versus-Set rationale** — the frozen-export test documents the guarantee, so the comment doesn't have to.

I read the rules as standing rather than specific to this file, so I've applied the same pass to the comments #414 adds. I'll do #415, #416, and #418 too unless you'd rather they wait until you've reviewed them.

---

packages/dropdown/src/Dropdown.tsx
acusti: Done in 9d99382, re-wrapped to the file's ~78 columns:

```js
    // A text input can’t serve as that source: aria-labelledby resolves one to
    // its *value*, so pointing at it would name the popup after whatever the
    // user has typed. That rules out triggers that are an input or textarea,
    // which have no accessible name. The fix is to pass props.label (or, for
    // the trigger alone, a custom trigger’s aria-label).
```

Two judgment calls, since the suggestion was clipped mid-sentence — flag either if I read it wrong:

- Filled "The is to pass" back in as "The fix is to pass".
- Kept the trailing "(or, for the trigger alone, a custom trigger's aria-label)" from the previous wording. `props.label` names both the trigger and the popup; `aria-label` on a custom trigger names only the trigger, so the parenthetical is the only place that distinction is recorded.

Your compression is right that naming the searchable case separately was redundant — its generated trigger *is* an `<input>`, so "triggers that are an input or textarea" already covers it, generated or custom. 134 dropdown tests still pass.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -173,8 +174,14 @@ const CHILDREN_ERROR =
     '@acusti/dropdown requires either 1 child (the dropdown body) or 2 children: the dropdown trigger and the dropdown body.';
 const CLICKABLE_SELECTOR = 'button, a[href], input[type="button"], input[type="submit"]';
 const FOCUSABLE_SELECTOR = 'a[href], button, input, select, textarea, [tabindex]';
-const TEXT_INPUT_SELECTOR =
-    'input:not([type=radio]):not([type=checkbox]):not([type=range]),textarea';
+// Any input that isn’t one of the non-text types, plus textarea. Derived from
+// the same list that drives use-keyboard-events’ isEventTargetUsingKeyEvent, so
+// the two can’t disagree about what counts as a text input — the hand-written
+// version of this listed only radio, checkbox, and range, which meant a submit
+// or button input in a custom trigger was adopted as its value source.
```
```suggestion
// the two can’t disagree about what counts as a text input.
```

acusti: Applied in e09a0fc — the two trailing lines about the old hand-written chain went with it:

```js
// Any input that isn’t one of the non-text types, plus textarea. Derived from
// the same list that drives use-keyboard-events’ isEventTargetUsingKeyEvent, so
// the two can’t disagree about what counts as a text input.
```

Agreed on the cut. What the previous selector missed is a fact about a version nobody will run, so it belongs in the changeset (where it is) rather than in a comment that has to be read forever. 134 tests pass; lint and format clean.

---

PR #414, opened 2026-08-04
Expose the active item to screen readers with aria-activedescendant

acusti:
**PR 4 of 6** in the pre-v1-RC a11y stack, and the one that motivated the rest. Base: #413 — review that first; this diff is only the newest commit.

## The gap

DOM focus stays on the trigger for as long as the dropdown is open — the highlight moves via `data-ukt-active`, not by moving focus — and nothing conveyed that to assistive technology. Verified before the fix, after opening a menu and pressing ↓:

```
active item html: <li data-ukt-value="a" role="menuitem" data-ukt-active="">Apple</li>
active item id:   (none)
trigger aria-activedescendant: null
document.activeElement: BUTTON uktdropdown-trigger
```

A screen reader announced the trigger and then went silent for every arrow key, typeahead jump, hover, and submenu dive — the whole keyboard model, invisible to the users who most depend on it.

## The approach

The trigger carries `aria-activedescendant` pointing at the highlighted item.

Rather than updating it at each of the ~12 `setActiveItem` call sites, `syncActiveDescendant` **re-derives it from the DOM** and is called from the three helpers that actually mutate active state (`setActiveChain`, `clearItemElementsState`, `collapseItem`) plus the two places in the component that mutate it directly — mousing out of an item, and closing, where the body unmounts and a stale id would dangle. A future caller can't forget it.

## Item ids

Extending the scheme from PR 3: the same `useId`, suffixed with the item's **index path** — its index at each level from the root down. `<bodyId>-2-1` is the second item of the submenu belonging to the third top-level item. Deterministic, collision-free, no module-level counter. A consumer-set `id` is left alone and pointed at as-is.

**Minted on activation, not in the annotation pass.** `aria-activedescendant` only ever needs an id for the active item, so on demand is sufficient — and it covers items rendered into an already-open body (async-loaded, or filtered by a consumer's own search), which the once-per-open pass doesn't reach and where an eager walk would leave `aria-activedescendant` pointing at nothing. There's a test for exactly that case.

Indices come from the navigable items, so a disabled item shifts the ids after it. Ids are minted and consumed within a single open and rewritten on every move, so nothing depends on their stability — noted in the code and the README.

## Incidental

`focusTrigger` and the new sync both need the trigger element, so that lookup moves into `helpers.ts` as `getTriggerElement`, and `FOCUSABLE_SELECTOR` moves with it.

## Tests

Nine new cases: highlight tracking, the index-path id shape (including `-2-1` in a submenu), following the highlight back out on collapse, clearing on close and on a failed typeahead, the searchable input, reveal-on-open, consumer-set ids, and the async-item case. All nine fail against the pre-fix source.

`bun run test` (137 passed), `lint`, `tsc`, and `format` all clean.

---

<a href="https://cubic.dev/pr/acusti/uikit/pull/414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -1324,10 +1324,12 @@ function RootDropdown({
         if (!ref) return;
         // Everything below must run exactly once per open: showPopover() throws
         // on an already-shown popover, and the ownership check would read the
-        // reveal’s own aria-selected annotation back as consumer-authored. A
-        // ref callback can re-attach on a re-render, and each open mounts a
-        // fresh body element, so latch per element (a WeakSet, to hold no
-        // unmounted bodies alive).
+        // reveal’s own aria-selected annotation back as consumer-authored.
+        // React re-attaches this ref to the same body element every time it
+        // re-renders with a new callback identity — three runs for a mount plus
+        // one prop change — so the latch is what makes “once per open” true.
+        // Keyed on the element, since each open mounts a fresh one, in a
+        // WeakSet so it holds no unmounted bodies alive.
         if (annotatedBodiesRef.current.has(ref)) return;
         annotatedBodiesRef.current.add(ref);
         // A consumer-authored aria-selected anywhere in the just-mounted body
@@ -1340,8 +1342,8 @@ function RootDropdown({
         annotateParentItems(ref);
         if (popupRole !== 'dialog') annotateItemRoles(ref, popupRole);
         // The Popover API is Baseline 2024, so showPopover() needs no feature
-        // detection; the body is mounted only while open, so this runs once
-        // per open and won’t throw on an already-open body.
+        // detection; the latch above is what keeps it off an already-shown
+        // popover, which would throw.
         ref.showPopover();
         // Reveal the current value on open: mark the matching option
         // aria-selected (listbox only — aria-selected isn’t valid on a
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -353,6 +353,17 @@ function RootDropdown({
         valueIdentity,
     ]);
 
+    // A consumer filtering or async-loading the body can drop the highlighted
+    // item, or shift it to a new index path, and nothing else would notice —
+    // the trigger would be left pointing at an id that has left the DOM or
+    // moved to another item. Deliberately dep-less: every render while open is
+    // a render that may have rearranged the items. It can’t hang off the body
+    // ref callback instead, because React Compiler memoizes that callback, so
+    // a re-render doesn’t re-attach the ref (see Dropdown.test.tsx).
+    useEffect(() => {
+        if (isOpen) syncActiveDescendant(dropdownElement);
+    });
+
     const isMountedRef = useRef(false);
 
     useEffect(() => {
@@ -1311,19 +1322,12 @@ function RootDropdown({
     // open. Known limitation until it’s needed.
     const handleBodyRef = (ref: HTMLDivElement | null) => {
         if (!ref) return;
-        // Re-running on every render while the body is open is exactly what
-        // this needs: a consumer filtering or async-loading the body can drop
-        // the highlighted item, and nothing else would notice — the trigger
-        // would be left pointing at an id that has left the DOM. Re-deriving
-        // here clears it (or moves it, if the item merely shifted position).
-        // above the once-per-open latch below, which would skip it
-        syncActiveDescendant(dropdownElement);
-        // An inline ref callback re-attaches (and re-runs) on every re-render
-        // while the body is open, but everything below must run exactly once
-        // per open: showPopover() throws on an already-shown popover, and the
-        // ownership check would read the reveal’s own aria-selected annotation
-        // back as consumer-authored. Each open mounts a fresh body element, so
-        // latch per element (a WeakSet, to hold no unmounted bodies alive).
+        // Everything below must run exactly once per open: showPopover() throws
+        // on an already-shown popover, and the ownership check would read the
+        // reveal’s own aria-selected annotation back as consumer-authored. A
+        // ref callback can re-attach on a re-render, and each open mounts a
+        // fresh body element, so latch per element (a WeakSet, to hold no
+        // unmounted bodies alive).
         if (annotatedBodiesRef.current.has(ref)) return;
         annotatedBodiesRef.current.add(ref);
         // A consumer-authored aria-selected anywhere in the just-mounted body
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -578,7 +578,7 @@ function RootDropdown({
     useEffect(() => clearHoverCloseTimer, []);
 
     const closeDropdown = (options?: { keepMenubarEngaged?: boolean }) => {
-        // The body (and its items) unmount on close, so the trigger would
+        // the body (and its items) unmount on close, so the trigger would
         // otherwise keep pointing at an id that no longer exists
         getTriggerElement(dropdownElement)?.removeAttribute('aria-activedescendant');
         setIsOpen(false);
@@ -1316,7 +1316,7 @@ function RootDropdown({
         // the highlighted item, and nothing else would notice — the trigger
         // would be left pointing at an id that has left the DOM. Re-deriving
         // here clears it (or moves it, if the item merely shifted position).
-        // Above the once-per-open latch below, which would skip it.
+        // above the once-per-open latch below, which would skip it
         syncActiveDescendant(dropdownElement);
         // An inline ref callback re-attaches (and re-runs) on every re-render
         // while the body is open, but everything below must run exactly once
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -44,10 +44,12 @@ import {
     getLevelRoot,
     getParentItem,
     getSubmenuOfItem,
+    getTriggerElement,
     isItemExpanded,
     isPointInTriangle,
     type Point,
     setActiveItem,
+    syncActiveDescendant,
 } from './helpers.js';
 
 export type Item = {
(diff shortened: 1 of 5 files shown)


acusti:
Heads up before you review this one: I found two of its own tests failing at its head, and the fix in afffe26c is a real behavior change, not a test tweak.

**What was wrong.** The re-derivation hung off the body ref callback, on the assumption that an inline ref re-attaches on every re-render. That's true for plain React, but **React Compiler memoizes the callback**, so its identity is stable and React doesn't re-attach the ref. A re-render that filtered or reordered items never re-derived, and the trigger kept pointing at a stale id — exactly what the feature exists to prevent.

**Why it wasn't caught earlier.** I'd had it backwards that the compiler applies only to the build. It's in `vite.config.base.js`'s top-level `plugins`, so vitest runs through it too. Confirmed empirically: adding a `console`-style statement inside `handleBodyRef` made the compiler bail out of memoizing it, the ref re-attached, and both tests passed — the instrumentation was creating the behavior it was measuring. With the instrumentation limited to `helpers.ts` (a plain module the compiler doesn't touch), the trace shows no `syncActiveDescendant` call at all after the re-render.

**The fix** moves it to a dep-less effect, which runs after every render regardless of memoization:

```js
useEffect(() => {
    if (isOpen) syncActiveDescendant(dropdownElement);
});
```

`handleBodyRef` keeps only its once-per-open work, which was already latched on a `WeakSet` and is unaffected.

Worth knowing more generally: **"an inline ref callback re-runs every render" is not a mechanism this codebase can rely on.** I grepped for other places leaning on it and found none, but it's the kind of assumption that reads as correct and silently isn't.

144 tests pass; lint, tsc, and format clean. #415, #416, and #418 still show these two failures because they're based on this PR's previous head — they'll clear when the stack restacks onto afffe26c.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -352,6 +353,17 @@ function RootDropdown({
...
+    // A consumer filtering or async-loading the body can drop the highlighted
+    // item, or shift it to a new index path, and nothing else would notice —
+    // the trigger would be left pointing at an id that has left the DOM or
+    // moved to another item. Deliberately dep-less: every render while open is
+    // a render that may have rearranged the items. It can’t hang off the body
+    // ref callback instead, because React Compiler memoizes that callback, so
+    // a re-render doesn’t re-attach the ref (see Dropdown.test.tsx).
+    useEffect(() => {
+        if (isOpen) syncActiveDescendant(dropdownElement);
+    });
```
acusti: “You might not need an effect” endorses using a conditional in render as an ok (though not ideal) alternative for state sync (with the conditional ensuring it prevents loops). that’s in reference to react state, which can be mutated synchronously during render without triggering another full render pass, but is that something we could leverage here also? maybe by moving this to directly before the `return` statement if needed? or should we use `useLayoutEffect` here to avoid double-paints?

acusti: `useLayoutEffect` is right, and it's in as 16ef474. The render-phase option can't work here, but for a reason worth being precise about.

**Why not render.** The "adjust state during render" escape hatch works because a setState during render is *deferred* — React discards the in-progress output and re-runs the component, and nothing has touched the DOM yet. That last part is exactly what breaks it for us: `syncActiveDescendant` doesn't set state, it reads the DOM (`getItemElements`, `getElementById`) and writes to it. During render the DOM still holds the **previous** render's children, so index paths would be derived from a body that no longer exists — the same bug in a new place. Measured it rather than assumed:

```
render(items=3): DOM has 0 li
  layoutEffect: DOM has 3 li
  effect:       DOM has 3 li
render(items=1): DOM has 3 li   ← the re-render sees the stale list
  layoutEffect: DOM has 1 li
  effect:       DOM has 1 li
```

Moving it just above the `return` doesn't change that — still pre-commit. And mutating the DOM in render breaks render purity generally: StrictMode's double-render would run the id-stripping twice, and SSR has no `document`.

**Why layout over passive.** Not double-paints, strictly — `aria-activedescendant` and the item `id` don't affect layout or pixels, so there's nothing to repaint. The real gap is that a passive effect runs *after* paint, leaving a window where the accessibility tree has the trigger pointing at an item that has left the document. Layout effects run in the commit, before the browser sees anything, so the id and the reference to it are always committed together. The cost is a handful of DOM reads and one `setAttribute` — nothing worth deferring past paint.

One thing I checked before committing, since the Menubar tab-stop reasoning leans on SSR: React 19 no longer warns about `useLayoutEffect` during server rendering. Confirmed on 19.2.8 — `renderToString` on a component using it produces the markup with zero `console.error` calls. So no `useIsomorphicLayoutEffect` shim needed, and this is the first `useLayoutEffect` in the repo.

Recorded the reasoning in the comment above the effect. 145 dropdown tests pass; tsc, lint, and format clean.

---

packages/dropdown/src/helpers.ts
```diff
@@ -318,21 +436,9 @@ export const collapseItemsOutsidePath = (
...
-        )) {
-            delete (active as HTMLElement).dataset.uktActive;
-        }
-    });
-};
-
-// Make the active set exactly the chain of element + its ancestor parent items
+// Make the active set exactly the chain of element + its ancestor parent items.
+// The only place data-ukt-active is added; clearActiveItems above is the only
+// place it’s removed.
```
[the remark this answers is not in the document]
acusti: Correct, and the comment was mine from an hour ago — fixed in 3349861.

`setActiveChain` does remove the attribute, on the items that fall outside the new chain:

```js
for (const active of getActiveItemElements(dropdownElement) ?? []) {
    if (!chain.has(active)) delete active.dataset.uktActive;
}
```

So "clearActiveItems is the only place it's removed" was false, and it's the kind of false invariant that reads as load-bearing. Both comments now say the true thing — that these are the **two** writers, and that both re-derive the reference before returning:

```js
// Clear the highlight on each scope element and on anything inside it. One of
// the two writers of data-ukt-active, alongside setActiveChain below.

// Make the active set exactly the chain of element + its ancestor parent items,
// clearing whatever falls outside it. The only place data-ukt-active is added,
// and the other of the two writers, alongside clearActiveItems above.
```

I went with that rather than the suggested "the only place that clears an entire active scope", since the property worth stating isn't which function clears the widest scope — it's that these two are the complete set of writers and each pairs its write with the sync. That's the invariant a future call site would otherwise break. 145 tests pass; lint and format clean.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -352,6 +355,24 @@ function RootDropdown({
...
+    // a render that may have rearranged the items. It can’t hang off the body
+    // ref callback instead, because React Compiler memoizes that callback, so
+    // a re-render doesn’t re-attach the ref (see Dropdown.test.tsx).
+    //
+    // Layout rather than passive so the id and the reference to it are always
+    // committed together: a passive effect would leave a window, however
+    // brief, where the accessibility tree has the trigger pointing at an item
+    // that is no longer in the document. It can’t move into render either —
+    // render runs before the commit, so it would read the previous render’s
+    // items and derive index paths from a body that no longer exists.
```
```suggestion
    // A consumer filtering or async-loading the body can drop or shift the highlighted
    // item without us seeing the change. Deliberately dep-less: every render while open
    // may have rearranged the items. useLayoutEffect over useEffect so the id and reference
    // to it are always committed together: a passive effect would leave a window where the a11y tree
    // has the trigger pointing at an item that is no longer in the document.
```
i’m sure it needs to be rewrapped

acusti: In as 9bfe070, your wording kept, wrapped to the file's ~78 columns:

```js
    // A consumer filtering or async-loading the body can drop or shift the
    // highlighted item without us seeing the change. Deliberately dep-less:
    // every render while open may have rearranged the items. useLayoutEffect
    // over useEffect so the id and reference to it are always committed
    // together: a passive effect would leave a window where the a11y tree has
    // the trigger pointing at an item that is no longer in the document.
```

15 lines to 6. The "can't move into render" paragraph went with it, which I agree doesn't earn its keep — nobody reaching for this code is going to try mutating the DOM during render.

One line I'd float putting back, and only because it already cost this PR a bug:

```js
    // Not the body ref callback: React Compiler memoizes it, so a re-render
    // doesn't re-attach the ref and the sync never runs.
```

That's where this logic lived until 90a7981, and the two filtering tests were failing against it. The trap is that the ref-callback version reads as obviously correct — inline refs do re-attach every render in plain React — so the next person optimizing this effect away has a good chance of landing right back on it. Your call; I've left it out rather than re-adding something you deliberately cut.

145 tests pass; lint and format clean.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1316,12 +1330,12 @@ function RootDropdown({
...
-        // while the body is open, but everything below must run exactly once
-        // per open: showPopover() throws on an already-shown popover, and the
-        // ownership check would read the reveal’s own aria-selected annotation
-        // back as consumer-authored. Each open mounts a fresh body element, so
-        // latch per element (a WeakSet, to hold no unmounted bodies alive).
+        // Everything below must run exactly once per open: showPopover() throws
+        // on an already-shown popover, and the ownership check would read the
+        // reveal’s own aria-selected annotation back as consumer-authored. A
+        // ref callback can re-attach on a re-render, and each open mounts a
+        // fresh body element, so latch per element (a WeakSet, to hold no
```
acusti: this might be my own issue, but the “A ref callback can re-attach on a re-render, ” addition doesn’t help me understand this (actually does the opposite)
```suggestion
        // reveal’s own aria-selected annotation back as consumer-authored.
        // Each open mounts a fresh body element, so latch per element (a WeakSet, to hold no
```
if it’s load-bearing, let’s try a different wording to see if it helps it land

acusti: Not your issue — the sentence was vague, and it turns out it was also underselling. Rewritten in 576a635 after measuring what actually happens.

It is load-bearing. I counted runs of the once-per-open block by instrumenting `annotateParentItems` (a plain module the compiler doesn't touch, so `handleBodyRef`'s memoization stays undisturbed), rendering open, then changing `value` — a dependency of the callback:

| | runs |
|---|--:|
| latch present | **1** |
| latch removed | **3** |

`sameBody=true` in both. React re-attaches the ref to the *same* body element each time the callback's identity changes, so the latch isn't guarding against a second open — it's what makes "once per open" true *within* one open. That's the part the old wording failed to say, and probably why it read backwards.

New version states the mechanism concretely:

```js
        // React re-attaches this ref to the same body element every time it
        // re-renders with a new callback identity — three runs for a mount plus
        // one prop change — so the latch is what makes “once per open” true.
        // Keyed on the element, since each open mounts a fresh one, in a
        // WeakSet so it holds no unmounted bodies alive.
```

While in there I fixed a second comment that was actively wrong, at the `showPopover()` call: *"the body is mounted only while open, so this runs once per open and won't throw on an already-open body."* The mounting isn't what protects it — the latch is. Without the latch that call runs three times, and a real browser throws on the second. happy-dom doesn't implement that check (verified: two `showPopover()` calls on the same element, no throw), which is why no test catches it.

145 tests pass; lint and format clean.

---

PR #415, opened 2026-08-04
Declare the searchable trigger a combobox

acusti:
**PR 5 of 6** in the pre-v1-RC a11y stack. Base: #414 — review that first; this diff is only the newest commit.

## The gap

The searchable trigger was documented as "the combobox pattern" (README:1092) but rendered as a bare text input:

```html
<input aria-controls="_r_1_" aria-expanded="false" aria-haspopup="listbox"
       autocomplete="off" class="uktdropdown-trigger" placeholder="search" type="text">
```

Without `role="combobox"` that input is exposed as a `textbox`, and ARIA 1.2 doesn't list `aria-expanded` among the states `textbox` supports — so the disclosure state the whole pattern turns on wasn't conveyed at all.

## The fix

`role="combobox"` plus `aria-autocomplete="list"`. The autocomplete value is `list` rather than `both` because typing here surfaces suggestions in the popup without completing anything inline — the input is only written back on submit.

A custom trigger gets the same treatment, gated on the trigger element **being** the text input, reusing the `isTextInputElement` check from PR 3. A trigger that merely *wraps* an input is left alone: `role="combobox"` on a wrapper is the deprecated ARIA 1.1 shape and would misdescribe the wrapper. A `role` or `aria-autocomplete` you set yourself still wins.

## ⚠️ Changes the trigger's exposed role

This is the one change in the stack that's visible to consumer code. Anything selecting the searchable trigger by role needs `combobox` instead of `textbox` — `getByRole` queries, `[role]` selectors. Non-searchable triggers are unaffected. Called out in both the changeset and the README.

In this repo that meant updating 29 test queries. The five left as `textbox` are genuinely still textboxes: one non-searchable custom input trigger, and inputs living inside dialog bodies.

## Tests

Five new cases: the generated combobox and its `aria-expanded` transition, a custom input trigger, the wrapper-not-labelled case, consumer overrides of `role`/`aria-autocomplete`, and a guard that non-searchable triggers stay plain buttons.

`bun run test` (142 passed), `lint`, `tsc`, and `format` all clean.

## One adjacent case I did *not* change

A **non-searchable** dropdown with a custom text-input trigger still gets `aria-expanded` + `aria-haspopup="menu"` on a plain textbox — the same class of invalid ARIA, in a rarer shape. I left it alone deliberately: `role="combobox"` isn't a clean fix there, because ARIA 1.2 allows a combobox's popup to be `listbox`/`tree`/`grid`/`dialog` but **not** `menu`, so it'd trade one invalid combination for another. Worth deciding separately — happy to take it in a follow-up.

---

<a href="https://cubic.dev/pr/acusti/uikit/pull/415?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/dropdown/README.md
+++ b/packages/dropdown/README.md
@@ -1100,16 +1100,19 @@ keyboard navigation. It works best when you use appropriate HTML elements
 in your dropdown content (like `<ul>` and `<li>` for lists, `<button>`
 elements for actions, etc.).
 
+The ARIA it adds is a best effort from what it can control: it annotates
+the DOM it renders and the shapes it recognizes, but never overwrites a
+value you pass. So when you provide a completely custom trigger, it is
+possible to end up with ARIA that doesn’t hold together. The safest thing
+if you go full custom is to take over the ARIA as well.
+
 ### ARIA attributes
 
 The trigger automatically receives `aria-haspopup`, `aria-expanded`, and
-`aria-controls` pointing to the open body. The exception is a custom
-trigger that is a text input or `<textarea>` and ends up with no `role`: it
-gets no `aria-expanded`, which `textbox` doesn’t support. A single-line
-`<input>` gets one back from `isSearchable`, which makes it a `combobox`
-(see below); a `<textarea>` — never a combobox, since that role is for
-single-line inputs — needs a `role` of its own. The popup role is chosen
-based on the dropdown’s mode:
+`aria-controls` pointing to the open body. The component does its best to
+also apply the correct roles and attributes with custom triggers, but if
+you are customizing in that way, consider also providing the appropriate
+ARIA. The popup role is chosen based on the dropdown’s mode:
 
 - `aria-haspopup="listbox"` when `isSearchable` is true (combobox pattern —
   see below)
@@ -1169,25 +1172,16 @@ ARIA you set. The persistent tint on the selected item is themeable via the
 ### The searchable trigger is a combobox
 
 A searchable dropdown’s trigger declares `role="combobox"` together with
-`aria-autocomplete="list"`. The role is what makes the rest of the pattern
-legal: ARIA 1.2 doesn’t include `aria-expanded` among the states a
-`textbox` supports, so a bare `<input>` carrying it — which is what this
-was before — is exposed without the disclosure state a combobox needs.
-`aria-autocomplete="list"` describes what typing actually does here:
+`aria-autocomplete="list"`, which describes what typing actually does here:
 suggestions appear in the popup, and nothing is completed inline in the
 input.
 
 For a custom trigger, the role is filled in only when the trigger element
-**is** a single-line text input. A trigger that merely wraps an input is
-left alone — `role="combobox"` on a wrapper is the deprecated ARIA 1.1
-shape, and it would misdescribe the wrapper — and a `<textarea>` is
-excluded for the same reason, since a combobox is single-line. (A textarea
-trigger still counts as a text input for naming, so it can’t name the popup
-either.) As always, a `role` or `aria-autocomplete` you set yourself wins.
-
-Note that this changes the trigger’s exposed role, so anything selecting it
-by role — `getByRole('textbox')` in tests, or a `[role]` selector — needs
-`combobox` for searchable dropdowns. Non-searchable triggers are unchanged.
+is a single-line text input. A trigger that merely wraps an input is left
+alone. Same with a `<textarea>`, since a combobox is single-line. (A
+textarea trigger still counts as a text input for naming, so it can’t name
+the popup either.) As always, a `role` or `aria-autocomplete` you set
+yourself wins.
 
 ### The active item and `aria-activedescendant`
 
--- a/packages/dropdown/README.md
+++ b/packages/dropdown/README.md
@@ -1106,7 +1106,8 @@ The trigger automatically receives `aria-haspopup`, `aria-expanded`, and
 `aria-controls` pointing to the open body. The popup role is chosen based
 on the dropdown’s mode:
 
-- `aria-haspopup="listbox"` when `isSearchable` is true (combobox pattern)
+- `aria-haspopup="listbox"` when `isSearchable` is true (combobox pattern —
+  see below)
 - `aria-haspopup="menu"` when `hasItems` is true (the default)
 - `aria-haspopup="dialog"` when `hasItems={false}` (interactive content)
 
@@ -1159,6 +1160,27 @@ on a `menuitem`) — unless you author `aria-selected` on items yourself, in
 which case the component leaves selection ARIA entirely to you, as with any
 ARIA you set. The persistent tint on the selected item is themeable via the
 `--uktdd-body-bg-color-selected` custom property.
+
+### The searchable trigger is a combobox
+
+A searchable dropdown’s trigger declares `role="combobox"` together with
+`aria-autocomplete="list"`. The role is what makes the rest of the pattern
+legal: ARIA 1.2 doesn’t include `aria-expanded` among the states a
+`textbox` supports, so a bare `<input>` carrying it — which is what this
+was before — is exposed without the disclosure state a combobox needs.
+`aria-autocomplete="list"` describes what typing actually does here:
+suggestions appear in the popup, and nothing is completed inline in the
+input.
+
+For a custom trigger, the role is filled in only when the trigger element
+**is** the text input. A trigger that merely wraps an input is left alone —
+`role="combobox"` on a wrapper is the deprecated ARIA 1.1 shape, and it
+would misdescribe the wrapper. As always, a `role` or `aria-autocomplete`
+you set yourself wins.
+
+Note that this changes the trigger’s exposed role, so anything selecting it
+by role — `getByRole('textbox')` in tests, or a `[role]` selector — needs
+`combobox` for searchable dropdowns. Non-searchable triggers are unchanged.
 
 ### The active item and `aria-activedescendant`
 
--- a/packages/dropdown/README.md
+++ b/packages/dropdown/README.md
@@ -1173,10 +1173,12 @@ suggestions appear in the popup, and nothing is completed inline in the
 input.
 
 For a custom trigger, the role is filled in only when the trigger element
-**is** the text input. A trigger that merely wraps an input is left alone —
-`role="combobox"` on a wrapper is the deprecated ARIA 1.1 shape, and it
-would misdescribe the wrapper. As always, a `role` or `aria-autocomplete`
-you set yourself wins.
+**is** a single-line text input. A trigger that merely wraps an input is
+left alone — `role="combobox"` on a wrapper is the deprecated ARIA 1.1
+shape, and it would misdescribe the wrapper — and a `<textarea>` is
+excluded for the same reason, since a combobox is single-line. (A textarea
+trigger still counts as a text input for naming, so it can’t name the popup
+either.) As always, a `role` or `aria-autocomplete` you set yourself wins.
 
 Note that this changes the trigger’s exposed role, so anything selecting it
 by role — `getByRole('textbox')` in tests, or a `[role]` selector — needs
--- a/packages/dropdown/README.md
+++ b/packages/dropdown/README.md
@@ -1103,8 +1103,13 @@ elements for actions, etc.).
 ### ARIA attributes
 
 The trigger automatically receives `aria-haspopup`, `aria-expanded`, and
-`aria-controls` pointing to the open body. The popup role is chosen based
-on the dropdown’s mode:
+`aria-controls` pointing to the open body. The exception is a custom
+trigger that is a text input or `<textarea>` and ends up with no `role`: it
+gets no `aria-expanded`, which `textbox` doesn’t support. A single-line
+`<input>` gets one back from `isSearchable`, which makes it a `combobox`
+(see below); a `<textarea>` — never a combobox, since that role is for
+single-line inputs — needs a `role` of its own. The popup role is chosen
+based on the dropdown’s mode:
 
 - `aria-haspopup="listbox"` when `isSearchable` is true (combobox pattern —
   see below)
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -1383,6 +1383,7 @@ function RootDropdown({
         if (isSearchable) {
             trigger = (
                 <input
+                    aria-autocomplete="list"
                     aria-controls={bodyId}
                     aria-expanded={isOpen}
                     aria-haspopup={popupRole}
@@ -1395,6 +1396,7 @@ function RootDropdown({
                     onFocus={() => setIsOpen(true)}
                     placeholder={placeholder}
                     ref={inputElementRef}
+                    role="combobox"
                     tabIndex={tabIndex}
                     type="text"
                 />
@@ -1425,15 +1427,24 @@ function RootDropdown({
         // point the body at whatever id the trigger ends up with, so a
         // consumer-set one is honored rather than overwritten
         const resolvedTriggerId = (triggerProps.id as string | undefined) ?? triggerId;
-        if (!isSearchable && !isTextInputElement(trigger)) {
+        const isTextInputTrigger = isTextInputElement(trigger);
+        if (!isSearchable && !isTextInputTrigger) {
             bodyLabelledBy = resolvedTriggerId;
         }
+        // role="combobox" belongs on the text input itself, so it’s filled in
+        // only when the custom trigger is one. A trigger that merely wraps an
+        // input is left alone: the role on the wrapper is the deprecated ARIA
+        // 1.1 shape, and putting it there would misdescribe the wrapper.
+        const isCombobox = isSearchable && isTextInputTrigger;
         trigger = cloneElement(trigger as ReactElement<Record<string, unknown>>, {
+            'aria-autocomplete':
+                triggerProps['aria-autocomplete'] ?? (isCombobox ? 'list' : undefined),
             'aria-controls': triggerProps['aria-controls'] ?? bodyId,
             'aria-disabled': triggerProps['aria-disabled'] ?? (disabled || undefined),
             'aria-expanded': triggerProps['aria-expanded'] ?? isOpen,
             'aria-haspopup': triggerProps['aria-haspopup'] ?? popupRole,
             id: resolvedTriggerId,
+            role: triggerProps.role ?? (isCombobox ? 'combobox' : undefined),
         });
     }
 
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -1436,15 +1436,20 @@ function RootDropdown({
         // textareas or non-input wrappers)
         const isCombobox =
             isSearchable && isTextInputTrigger && trigger.type !== 'textarea';
+        const triggerRole = triggerProps.role ?? (isCombobox ? 'combobox' : undefined);
+        // aria-expanded isn’t supported on textbox either, so a text input
+        // only gets one once a role makes it something else
+        const canExpand = !isTextInputTrigger || triggerRole != null;
         trigger = cloneElement(trigger as ReactElement<Record<string, unknown>>, {
             'aria-autocomplete':
                 triggerProps['aria-autocomplete'] ?? (isCombobox ? 'list' : undefined),
             'aria-controls': triggerProps['aria-controls'] ?? bodyId,
             'aria-disabled': triggerProps['aria-disabled'] ?? (disabled || undefined),
-            'aria-expanded': triggerProps['aria-expanded'] ?? isOpen,
+            'aria-expanded':
+                triggerProps['aria-expanded'] ?? (canExpand ? isOpen : undefined),
             'aria-haspopup': triggerProps['aria-haspopup'] ?? popupRole,
             id: resolvedTriggerId,
-            role: triggerProps.role ?? (isCombobox ? 'combobox' : undefined),
+            role: triggerRole,
         });
     }
 
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -1431,11 +1431,11 @@ function RootDropdown({
         if (!isSearchable && !isTextInputTrigger) {
             bodyLabelledBy = resolvedTriggerId;
         }
-        // role="combobox" belongs on the text input itself, so it’s filled in
-        // only when the custom trigger is one. A trigger that merely wraps an
-        // input is left alone: the role on the wrapper is the deprecated ARIA
-        // 1.1 shape, and putting it there would misdescribe the wrapper.
-        const isCombobox = isSearchable && isTextInputTrigger;
+        // role="combobox" & aria-autocomplete="list" belong on the text input,
+        // but only when the trigger is an actual text input (not valid on
+        // textareas or non-input wrappers)
+        const isCombobox =
+            isSearchable && isTextInputTrigger && trigger.type !== 'textarea';
         trigger = cloneElement(trigger as ReactElement<Record<string, unknown>>, {
             'aria-autocomplete':
                 triggerProps['aria-autocomplete'] ?? (isCombobox ? 'list' : undefined),
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -1319,7 +1319,7 @@ function RootDropdown({
     // These annotations run once per open (the body unmounts on close and is
     // annotated fresh each time); items rendered into an already-open body —
     // async-loaded or consumer-filtered — aren’t annotated until the next
-    // open. Known limitation until it’s needed.
+    // open. Custom implementations like that should set their own ARIA roles.
     const handleBodyRef = (ref: HTMLDivElement | null) => {
         if (!ref) return;
         // Everything below must run exactly once per open: showPopover() throws
--- a/.changeset/dropdown-combobox-role.md
+++ b/.changeset/dropdown-combobox-role.md
@@ -1,0 +1,17 @@
+---
+'@acusti/dropdown': minor
+---
+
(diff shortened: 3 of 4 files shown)


packages/dropdown/src/Dropdown.tsx
```diff
@@ -1425,15 +1427,30 @@ function RootDropdown({
...
         trigger = cloneElement(trigger as ReactElement<Record<string, unknown>>, {
+            'aria-autocomplete': isCombobox
+                ? (triggerProps['aria-autocomplete'] ?? 'list')
+                : triggerProps['aria-autocomplete'],
             'aria-controls': triggerProps['aria-controls'] ?? bodyId,
             'aria-disabled': triggerProps['aria-disabled'] ?? (disabled || undefined),
             'aria-expanded': triggerProps['aria-expanded'] ?? isOpen,
             'aria-haspopup': triggerProps['aria-haspopup'] ?? popupRole,
             id: resolvedTriggerId,
+            role: isCombobox ? (triggerProps.role ?? 'combobox') : triggerProps.role,
```
acusti: both of these new ternaries would read more clearly to me (and better match the existing examples) if they start with the `triggerProps`:
```suggestion
            'aria-autocomplete': triggerProps['aria-autocomplete'] ?? (isCombobox ? 'list' : undefined),
            'aria-controls': triggerProps['aria-controls'] ?? bodyId,
            'aria-disabled': triggerProps['aria-disabled'] ?? (disabled || undefined),
            'aria-expanded': triggerProps['aria-expanded'] ?? isOpen,
            'aria-haspopup': triggerProps['aria-haspopup'] ?? popupRole,
            id: resolvedTriggerId,
            role: triggerProps.role ?? (isCombobox ? 'combobox' : undefined),
```
(subject to `oxfmt`)

acusti: Agreed, and in as 0d2abcb. oxfmt wrapped the first one:

```js
            'aria-autocomplete':
                triggerProps['aria-autocomplete'] ?? (isCombobox ? 'list' : undefined),
            'aria-controls': triggerProps['aria-controls'] ?? bodyId,
            'aria-disabled': triggerProps['aria-disabled'] ?? (disabled || undefined),
            'aria-expanded': triggerProps['aria-expanded'] ?? isOpen,
            'aria-haspopup': triggerProps['aria-haspopup'] ?? popupRole,
            id: resolvedTriggerId,
            role: triggerProps.role ?? (isCombobox ? 'combobox' : undefined),
```

Now all seven read the same way — consumer's value first, component default as the fallback — instead of two of them burying that behind the `isCombobox` check.

Behavior is unchanged: the old form's non-combobox branch already fell through to `triggerProps[…]`, so a consumer value survived either way. The only difference is an explicit `null` from a consumer now becomes `undefined`, which React treats identically for an attribute.

151 tests pass; tsc, lint, and format clean.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1425,15 +1427,30 @@ function RootDropdown({
...
             bodyLabelledBy = resolvedTriggerId;
         }
+        // role="combobox" belongs on the text input itself, so it’s filled in
+        // only when the custom trigger is one. A trigger that merely wraps an
+        // input is left alone: the role on the wrapper is the deprecated ARIA
+        // 1.1 shape, and putting it there would misdescribe the wrapper. A
+        // textarea is excluded for the same reason — a combobox is a
+        // single-line input, so the role would misdescribe a multi-line field
+        // — though it still counts as a text input for naming, where what
+        // matters is that aria-labelledby resolves it to its value.
```
acusti: i don’t think we need to bring up naming and `aria-labelledby` here
```suggestion
        // role="combobox" & aria-autocomplete="list" belong on the text input, but only
        // when the trigger is an actual text input (not valid on textareas or non-input wrappers)
```

acusti: Fair — the naming half belongs at `isTextInputElement`, not here. In as 111c8b1, re-wrapped to ~78:

```js
        // role="combobox" & aria-autocomplete="list" belong on the text input,
        // but only when the trigger is an actual text input (not valid on
        // textareas or non-input wrappers)
```

8 lines to 3. Your version also covers `aria-autocomplete`, which the old one never mentioned despite deciding it on the same line.

The ARIA 1.1 detail that went with it isn't lost — it's still where someone would look for it:

- `Dropdown.test.tsx:190` — `// role=combobox belongs on the input itself; putting it on a wrapper is the deprecated ARIA 1.1 shape`
- `README.md:1177` — same point in the accessibility section

151 tests pass; lint and format clean.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1425,15 +1427,24 @@ function RootDropdown({
...
+        // textareas or non-input wrappers)
+        const isCombobox =
+            isSearchable && isTextInputTrigger && trigger.type !== 'textarea';
         trigger = cloneElement(trigger as ReactElement<Record<string, unknown>>, {
+            'aria-autocomplete':
+                triggerProps['aria-autocomplete'] ?? (isCombobox ? 'list' : undefined),
             'aria-controls': triggerProps['aria-controls'] ?? bodyId,
             'aria-disabled': triggerProps['aria-disabled'] ?? (disabled || undefined),
             'aria-expanded': triggerProps['aria-expanded'] ?? isOpen,
             'aria-haspopup': triggerProps['aria-haspopup'] ?? popupRole,
```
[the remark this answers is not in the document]
acusti: Valid. Took option 2, in c5a73e3 — but the scope is different from the comment's in both directions, so it's worth laying out what I measured.

I rendered every custom-trigger shape and dumped what each ends up with:

```
searchable + textarea:        <textarea> aria-controls aria-expanded aria-haspopup="listbox"
searchable + input[text]:     <input>    aria-controls aria-expanded aria-haspopup="listbox" aria-autocomplete role="combobox"
NOT searchable + input[text]: <input>    aria-controls aria-expanded aria-haspopup="menu"
NOT searchable + div wrapper: <div>      aria-controls aria-expanded aria-haspopup="menu"
NOT searchable + button:      <button>   aria-controls aria-expanded aria-haspopup="menu"
```

**Narrower than stated on one axis:** `aria-haspopup` *is* supported on `textbox` — per the ARIA attribute's role list it's used in `application`, `button`, `combobox`, `gridcell`, `link`, `menuitem`, `slider`, **`textbox`**, `treeitem`. So "possibly other popup attributes" doesn't apply to it; only `aria-expanded` is the problem, whose list is `application`, `button`, `checkbox`, `combobox`, `gridcell`, `link`, `listbox`, `menuitem`, `row`, `rowheader`, `tab`, `treeitem` — no `textbox`.

**Broader on another:** it isn't specific to `isSearchable` + textarea. Row 3 above — a **non-searchable** dropdown with a custom text `<input>` trigger — is the same invalid pairing, and predates this PR.

So the fix keys on the outcome rather than on the shape:

```js
// aria-expanded isn’t supported on textbox either, so a text input
// trigger only gets one once the role above makes it a combobox
const canExpand = isCombobox || !isTextInputTrigger;
```

Both textarea and non-searchable text inputs now go without, `aria-haspopup` stays on all of them, and a consumer-set `aria-expanded` still wins.

I skipped option 1 (dev warning) because the shape isn't misuse — a text input trigger is a reasonable thing to pass, and it works; only that one attribute was wrong.

**One gap left, deliberately.** Row 4 — a `<div>` wrapper trigger — carries `aria-expanded` *and* `aria-haspopup` on a `generic`, which supports neither. That isn't fixable from here: a custom trigger can be any component rendering any element, so there's no reliable way to know its role. Text inputs are detectable, which is why they're covered. Every custom-trigger example in the README is a `<button>`, which takes both correctly.

Three tests added, two of which fail against the previous commit. README and changeset note the exception. 153 tests pass; tsc, lint, and format clean.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1425,15 +1427,28 @@ function RootDropdown({
...
             bodyLabelledBy = resolvedTriggerId;
         }
+        // role="combobox" & aria-autocomplete="list" belong on the text input,
+        // but only when the trigger is an actual text input (not valid on
+        // textareas or non-input wrappers)
+        const isCombobox =
+            isSearchable && isTextInputTrigger && trigger.type !== 'textarea';
+        // aria-expanded isn’t supported on textbox either, so a text input
+        // trigger only gets one once the role above makes it a combobox
+        const canExpand = isCombobox || !isTextInputTrigger;
```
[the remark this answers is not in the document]
acusti: Valid, and the general form turned out better than the special case. Fixed in 09b1d24.

Instead of adding `triggerProps.role === 'combobox'` as a third disjunct, I hoisted the role the element actually ends up with and derived the answer from it:

```js
const triggerRole = triggerProps.role ?? (isCombobox ? 'combobox' : undefined);
// aria-expanded isn’t supported on textbox either, so a text input
// trigger only gets one once a role — ours or the consumer’s — makes
// it something other than one
const canExpand = !isTextInputTrigger || triggerRole != null;
```

Same result for the case you raised, and it also covers a consumer declaring some other expandable role (`role="button"` on an input, say) rather than only `combobox`. It drops a duplicated expression too — `role:` in the `cloneElement` is now just `triggerRole`.

The shape it doesn't cover is a consumer writing `role="textbox"` explicitly on a text input, which would then carry an `aria-expanded` it can't use. That's a redundant role to write, and treating an explicit role as the consumer's declaration of intent is how the rest of these props already behave.

Retargeted the test I'd added: it now sets **only** `role="combobox"`, no `aria-expanded`, and asserts the attribute comes back. It fails against the previous commit.

153 tests pass; tsc, lint, and format clean.

---

packages/dropdown/README.md
```diff
@@ -1103,10 +1103,14 @@ elements for actions, etc.).
...
 
 The trigger automatically receives `aria-haspopup`, `aria-expanded`, and
-`aria-controls` pointing to the open body. The popup role is chosen based
-on the dropdown’s mode:
-
-- `aria-haspopup="listbox"` when `isSearchable` is true (combobox pattern)
+`aria-controls` pointing to the open body. The exception is a custom
+trigger that is a text input or `<textarea>` with no `role` of its own: it
+gets no `aria-expanded`, which `textbox` doesn’t support. Giving it a
+`role` — or `isSearchable`, which makes it a `combobox` (see below) —
```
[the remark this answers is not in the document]
acusti: Right — `isSearchable` doesn't reach a `<textarea>`, since `isCombobox` excludes it by design. Fixed in 5c64524.

Verified rather than reasoned about, by rendering that exact shape:

```
searchable + <textarea>:  role=(none)  aria-expanded=(none)  aria-haspopup=listbox
```

No role, so no `aria-expanded` — the sentence promised one. New wording splits the two cases instead of lumping them:

> A single-line `<input>` gets one back from `isSearchable`, which makes it a `combobox` (see below); a `<textarea>` — never a combobox, since that role is for single-line inputs — needs a `role` of its own.

This is the third documentation error you've caught on this change, all in the same paragraph and all mine: first the changeset claiming wrappers lose `aria-expanded`, then the code comment, now this. The implementation has held up each time; the prose describing it hasn't. Fair signal to distrust my summaries of this block more than the block itself.

153 tests pass; format clean. #416 and #418 restacked onto this (`f16c209f`, `7efc3fb2`), 163 passing at the top.

---

packages/dropdown/README.md
```diff
@@ -1100,13 +1100,29 @@ keyboard navigation. It works best when you use appropriate HTML elements
...
+`aria-controls` pointing to the open body. The exception is a custom
+trigger that is a text input or `<textarea>` and ends up with no `role`: it
+gets no `aria-expanded`, which `textbox` doesn’t support. A single-line
+`<input>` gets one back from `isSearchable`, which makes it a `combobox`
+(see below); a `<textarea>` — never a combobox, since that role is for
+single-line inputs — needs a `role` of its own. The popup role is chosen
+based on the dropdown’s mode:
+
+- `aria-haspopup="listbox"` when `isSearchable` is true (combobox pattern —
+  see below)
```
acusti: > A single-line `<input>` gets one back from `isSearchable`, which makes it a `combobox`

is that saying that you can get a `role` by passing `isSearchable` as `true`? or something else? if the former, we could say something like “A single-line `<input>` gets `role="combobox"` when `isSearchable`.” but the “(see below)” points out that even that mention is redundant with the new “The searchable trigger is a combobox” section.

maybe we should simplify to something like:

```suggestion
`aria-controls` pointing to the open body. The component does its best
to also apply the correct roles and attributes with custom triggers, but
if you are customizing in that way, the safest thing is to also provide
the appropriate ARIA as a part of that customization. The popup role
is chosen based on the dropdown’s mode:

- `aria-haspopup="listbox"` when `isSearchable` is true (combobox pattern —
  see below)
```

packages/dropdown/README.md
```diff
@@ -1100,13 +1100,29 @@ keyboard navigation. It works best when you use appropriate HTML elements
...
 
+The ARIA it adds is a best effort from what it can see: it annotates the
+shapes it recognizes and never overwrites a value you pass. Where it can’t
+tell what you meant — a custom trigger can be any element rendering
+anything — it leaves your markup alone rather than guessing. So it is
+possible to end up with ARIA that doesn’t hold together: a `<div>` trigger
+carrying an `aria-expanded` its role can’t take, or a text input nested in
+a custom trigger naming the popup after whatever was typed into it. That’s
+the trade — flexibility first, and the last word on your own markup is
+yours.
```
```suggestion
The ARIA it adds is a best effort from what it can control: it annotates
the DOM it renders and the shapes it recognizes, but never overwrites
a value you pass. So when you provide a completely custom trigger, it is
possible to end up with ARIA that doesn’t hold together. The safest thing
if you go full custom is to take over the ARIA as well.
```

packages/dropdown/README.md
```diff
@@ -1160,6 +1176,29 @@ which case the component leaves selection ARIA entirely to you, as with any
...
+### The searchable trigger is a combobox
+
+A searchable dropdown’s trigger declares `role="combobox"` together with
+`aria-autocomplete="list"`. The role is what makes the rest of the pattern
+legal: ARIA 1.2 doesn’t include `aria-expanded` among the states a
+`textbox` supports, so a bare `<input>` carrying it — which is what this
+was before — is exposed without the disclosure state a combobox needs.
+`aria-autocomplete="list"` describes what typing actually does here:
+suggestions appear in the popup, and nothing is completed inline in the
+input.
```
```suggestion
`aria-autocomplete="list"`, which describes what typing actually does here:
suggestions appear in the popup, and nothing is completed inline in the
input.
```

packages/dropdown/README.md
```diff
@@ -1160,6 +1176,29 @@ which case the component leaves selection ARIA entirely to you, as with any
...
+
+For a custom trigger, the role is filled in only when the trigger element
+**is** a single-line text input. A trigger that merely wraps an input is
+left alone — `role="combobox"` on a wrapper is the deprecated ARIA 1.1
+shape, and it would misdescribe the wrapper — and a `<textarea>` is
+excluded for the same reason, since a combobox is single-line. (A textarea
```
```suggestion
is a single-line text input. A trigger that merely wraps an input is left
alone. Same with a `<textarea>`, since a combobox is single-line. (A textarea
```

packages/dropdown/README.md
```diff
@@ -1160,6 +1176,29 @@ which case the component leaves selection ARIA entirely to you, as with any
...
+trigger still counts as a text input for naming, so it can’t name the popup
+either.) As always, a `role` or `aria-autocomplete` you set yourself wins.
+
+Note that this changes the trigger’s exposed role, so anything selecting it
+by role — `getByRole('textbox')` in tests, or a `[role]` selector — needs
+`combobox` for searchable dropdowns. Non-searchable triggers are unchanged.
+
```
acusti: this is changelog info, not readme info
```suggestion
```

packages/dropdown/README.md
acusti: All five in as f03d573, README down 29 lines and up 18.

To answer the question first: yes, the former — `isSearchable` is what makes a single-line `<input>` a `combobox`, so it was saying you get a role by setting a different prop. And you're right that the `(see below)` gave it away: if the reader has to go to the combobox section anyway, the enumeration here was just a worse copy of it. Your version says the thing that's actually load-bearing for a consumer — if you go custom, own the ARIA — and lets the section below carry the specifics.

That also retires a paragraph I'd rewritten twice under review (once for wrappers, once for textareas). It was a bad shape: an exhaustive list of which trigger shape ends up with which attribute is only correct until the next branch is added, and I got it wrong both times. One policy sentence can't drift like that.

The other four are in as written, with two judgment calls:

- **`aria-autocomplete`**: your text starts with `` `aria-autocomplete="list"`, which describes… ``, which left a fragment where the old sentence stood alone. I made it *"It pairs with `aria-autocomplete="list"`, which describes what typing actually does here: …"*, attaching to the sentence before it. Say the word if you meant something else.
- **`**is**` → `is`**: your suggestion dropped the bold, so I took it as intentional and left it plain.

The ARIA 1.1 wrapper rationale is still in `Dropdown.test.tsx:190`, and the exposed-role upgrade note is still in the changeset, so both survive where they belong.

153 tests pass; lint and format clean. #416 and #418 restacked (`120cd09f`, `e2d3fd5d`), 163 passing at the top.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1319,7 +1319,8 @@ function RootDropdown({
     // These annotations run once per open (the body unmounts on close and is
     // annotated fresh each time); items rendered into an already-open body —
     // async-loaded or consumer-filtered — aren’t annotated until the next
-    // open. Known limitation until it’s needed.
+    // open. Set their roles yourself to cover that: a consumer-set role is
+    // always left alone.
```
```suggestion
    // open. Custom implementations like that should set their own ARIA roles.
```

acusti: In verbatim as 83ff552 — two lines down to one:

```js
    // These annotations run once per open (the body unmounts on close and is
    // annotated fresh each time); items rendered into an already-open body —
    // async-loaded or consumer-filtered — aren’t annotated until the next
    // open. Custom implementations like that should set their own ARIA roles.
```

Reads better than what I had (*"Set their roles yourself to cover that: a consumer-set role is always left alone"*), which was explaining the mechanism when the reader only needs to know whose job it is.

---

packages/dropdown/README.md
```diff
@@ -1100,13 +1100,23 @@ keyboard navigation. It works best when you use appropriate HTML elements
...
+if you go full custom is to take over the ARIA as well.
+
 ### ARIA attributes
 
 The trigger automatically receives `aria-haspopup`, `aria-expanded`, and
-`aria-controls` pointing to the open body. The popup role is chosen based
-on the dropdown’s mode:
-
-- `aria-haspopup="listbox"` when `isSearchable` is true (combobox pattern)
+`aria-controls` pointing to the open body. The component does its best to
```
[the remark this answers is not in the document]
acusti: Valid — the two paragraphs landed four lines apart and say the same thing. Deduped in de0f138, keeping the intro as suggested:

```md
### ARIA attributes

The trigger automatically receives `aria-haspopup`, `aria-expanded`, and
`aria-controls` pointing to the open body. The popup role is chosen based
on the dropdown’s mode:
```

@acusti — flagging that this removes a sentence you wrote yourself, in the suggestion on the `### ARIA attributes` paragraph. The two suggestions were written a few minutes apart and I applied both, so the collision is on me for not reading them together. I kept the one in the section intro because it carries the "here's the trade" framing and lands before a reader hits any of the specifics; the `### ARIA attributes` copy was the same advice with no new information. Trivial to restore if you'd rather have it at both altitudes.

153 tests pass; format clean. Restacked: #416 `7043b533`, #418 `bd09aadb`.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1425,15 +1428,30 @@ function RootDropdown({
...
         }
+        // role="combobox" & aria-autocomplete="list" belong on the text input,
+        // but only when the trigger is an actual text input (not valid on
+        // textareas or non-input wrappers)
+        const isCombobox =
+            isSearchable && isTextInputTrigger && trigger.type !== 'textarea';
+        const triggerRole = triggerProps.role ?? (isCombobox ? 'combobox' : undefined);
+        // aria-expanded isn’t supported on textbox either, so a text input
+        // trigger only gets one once a role — ours or the consumer’s — makes
+        // it something other than one
```
acusti: this is hard to follow. i think the gist is this?
```suggestion
        // aria-expanded isn’t supported on textbox either, so a text input
        // only gets one if its role supports it
```

acusti: Yes, that's the gist — mine was doing too much work. In as aa193bd, with one phrase changed:

```js
        // aria-expanded isn’t supported on textbox either, so a text input
        // only gets one once a role makes it something else
        const canExpand = !isTextInputTrigger || triggerRole != null;
```

"if its role supports it" would describe a check the code doesn't do: `canExpand` is `triggerRole != null`, so it only asks whether a role is *there*, not which one. Whatever role you declare is taken at face value, same as everything else in this block. "once a role makes it something else" keeps your length and shape without implying validation.

153 tests pass; lint and format clean. Restacked: #416 `fb33bc39`, #418 `a077b4f5`.

---

packages/dropdown/README.md
```diff
@@ -1100,13 +1100,23 @@ keyboard navigation. It works best when you use appropriate HTML elements
...
+also apply the correct roles and attributes with custom triggers, but if
+you are customizing in that way, the safest thing is to also provide the
+appropriate ARIA as a part of that customization. The popup role is chosen
+based on the dropdown’s mode:
```
acusti: just realized that this is reiterating literally what was said in the previous paragraph (i didn’t appreciate that they were sibling paragraphs in my last review pass). we can simplify further 
```suggestion
`aria-controls` pointing to the open body. The component does its best to
also apply the correct roles and attributes with custom triggers, but if you
are customizing in that way, consider also providing the appropriate ARIA.
The popup role is chosen based on the dropdown’s mode:
```

acusti: You and cubic caught the same thing within a minute of each other, and I'd already acted on cubic's version — it suggested deleting the sentence outright, which I did in de0f138, a few minutes before this came in. So you're reviewing f03d573 and the line is already gone at HEAD.

Your resolution is different from cubic's, though: you wrote replacement prose rather than an empty suggestion, and you've used empty suggestions in this same pass when you wanted something deleted. So I read this as wanting a softened pointer kept, not a removal, and restored it in your words (d8755c0):

```md
The trigger automatically receives `aria-haspopup`, `aria-expanded`, and
`aria-controls` pointing to the open body. The component does its best to
also apply the correct roles and attributes with custom triggers, but if
you are customizing in that way, consider also providing the appropriate
ARIA. The popup role is chosen based on the dropdown’s mode:
```

Which lands better than either of the two previous states, for what it's worth: the intro says *the safest thing is to take over the ARIA*, and this now says *consider also providing it* — a pointer at the point of use rather than the same rule stated twice at the same strength.

Net effect is that de0f138 and d8755c0 mostly cancel out. Happy to squash them if you'd rather the history not show the round trip; leaving them separate for now since each is a real response to a distinct review comment.

153 tests pass; format clean. Restacked: #416 `02947bfe`, #418 `01242db7`.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1425,15 +1427,29 @@ function RootDropdown({
...
             bodyLabelledBy = resolvedTriggerId;
         }
+        // role="combobox" & aria-autocomplete="list" belong on the text input,
+        // but only when the trigger is an actual text input (not valid on
+        // textareas or non-input wrappers)
+        const isCombobox =
+            isSearchable && isTextInputTrigger && trigger.type !== 'textarea';
+        const triggerRole = triggerProps.role ?? (isCombobox ? 'combobox' : undefined);
+        // aria-expanded isn’t supported on textbox either, so a text input
+        // only gets one once a role makes it something else
```
acusti, line 1441: should we do `|| (triggerRole != null && triggerRole !== 'textbox')` to make this fully accurate?

acusti: It'd be more accurate, but not fully — and the gap it leaves is the reason I'd skip it.

`textbox` isn't the only role a text input can plausibly carry that doesn't support `aria-expanded`. From the attribute's role list (used in `application`, `button`, `checkbox`, `combobox`, `gridcell`, `link`, `listbox`, `menuitem`, `row`, `rowheader`, `tab`, `treeitem`; inherits into `columnheader`, `menuitemcheckbox`, `menuitemradio`, `rowheader`, `switch`), two others are equally unsupported and equally reachable here:

- `role="searchbox"` on `<input type="search">` — a subclass of textbox, and a natural thing to write
- `role="spinbutton"` on `<input type="number">` — `number` isn't in `NON_TEXT_INPUT_TYPES`, so `isTextInputElement` counts it as a text input

So `!== 'textbox'` fixes one instance and starts a list, and a partial blocklist reads as authoritative while missing its neighbours. That's the same shape that went wrong three times in the README paragraph enumerating trigger types.

Two coherent options:

1. **Leave it** — a declared role is taken at face value, like every other prop in this block. Matches where we landed on best-effort, and reaching the bad state means writing a redundant role on your own input.
2. **Invert to an allowlist** — `EXPANDABLE_ROLES.has(triggerRole)` over the ~17 roles above. Genuinely fully accurate, and it can't drift since the list is spec-stable. The cost is that it overrides a role you declared, which is the one thing this block currently never does.

I'd keep 1, on the reasoning that we stop guessing once you've told us what the element is. But 2 is about six lines and a test if you'd rather be strict — say which and I'll do it.

---

acusti: for the record, we went with option 1 (“Leave it”)

PR #416, opened 2026-08-04
Give Menubar real menubar semantics

acusti:
**PR 6 of 6** in the pre-v1-RC a11y stack. Base: #415 — review that first; this diff is only the newest commit.

## The gap

`Menubar` rendered `role="menubar"` but owned nothing a menubar can own:

```html
<div class="uktmenubar" role="menubar">
  <div class="uktdropdown"><label…><button aria-haspopup="menu" tabindex="0">File</button>
```

A menubar's required owned elements are menu items. The generic `div` (and `<label>`, with `props.label`) sat in between and the trigger stayed a plain button, so the bar was reported as having none — the `aria-required-children` failure. Every trigger was also its own tab stop, where APG expects the whole bar to be one.

## The fix

Triggers take `role="menuitem"`; the elements between them and the bar are neutralized with `role="none"`. A **searchable member is exempt** — its trigger is a `combobox`, which isn't a valid menubar child, so it keeps its own semantics rather than being forced into a menu item it isn't.

The roving tabindex is held as state on the bar, since only the bar knows document order. It follows the trigger you last focused, and starts on — and falls back to — the first enabled member, so a disabled or unmounted holder can't leave the bar unreachable.

## Two subtleties worth flagging for review

Both cost me a debugging pass, and both are commented in place:

1. **Registration claims the tab stop itself.** Members register from their own effects, which don't re-render `Menubar`, so the reconciliation effect would never see them arrive. Members register in document order, so the first enabled one gets it.

2. **Registration cleanup deliberately does *not* release the tab stop.** Members re-register on every render, and React runs every cleanup before any setup — so releasing handed the stop to whichever member re-registered first (always the first in the bar) rather than back to its holder. The symptom was that focusing any trigger but the first appeared to do nothing. A genuinely unmounted holder is caught by the reconciliation effect instead, which no longer finds it among the members.

The reconciliation effect intentionally runs on every render rather than on a dependency list — an unmounting member changes no state to depend on. The `exhaustive-deps` suppression carries that reasoning, plus why the setState can't chain.

## ⚠️ Changes the triggers' exposed role

Like #415, this is visible to consumer code: selecting a menubar member's trigger by role needs `menuitem` rather than `button`. Standalone dropdowns outside a `Menubar` are unaffected. The Menubar tests are updated accordingly; the one plain `<button>` in those fixtures is a non-menu control and stays a button.

## Tests

Six new cases under `describe('menubar semantics')`, including a structural one that walks from each menuitem up to the bar asserting every ancestor is `role="none"` — that's the `aria-required-children` condition expressed directly.

`bun run test` (148 passed), `lint`, `tsc`, and `format` all clean.

---

<a href="https://cubic.dev/pr/acusti/uikit/pull/416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -297,6 +297,19 @@ function RootDropdown({
     const labelId = `${bodyId}-label`;
     const popupRole = isSearchable ? 'listbox' : hasItems ? 'menu' : 'dialog';
     const menubar = useContext(MenubarContext);
+    // A menubar owns menuitems, so a member's trigger takes role="menuitem"
+    // and the elements between the bar and it are neutralized — a generic div
+    // (or label) in between leaves the menubar owning no menuitems at all.
+    // Only a menu trigger qualifies: a searchable member is a combobox, which
+    // isn't a valid menubar child, so it keeps its own semantics.
+    const isMenubarItem = menubar != null && popupRole === 'menu';
+    // The bar keeps a single tab stop (APG roving tabindex); the members that
+    // don't hold it stay reachable with ←/→
+    const menubarTabIndex = !isMenubarItem
+        ? undefined
+        : menubar.tabbableElement === dropdownElement
+          ? 0
+          : -1;
     const inputElementRef = useRef<HTMLInputElement | null>(null);
     // Whether the consumer authors aria-selected in the body themselves, in
     // which case this component’s aria-selected fill-ins stand down (set per
@@ -1411,7 +1424,8 @@ function RootDropdown({
                     className="uktdropdown-trigger"
                     disabled={disabled}
                     id={triggerId}
-                    tabIndex={0}
+                    role={isMenubarItem ? 'menuitem' : undefined}
+                    tabIndex={menubarTabIndex ?? 0}
                     type="button"
                 >
                     {trigger}
@@ -1436,7 +1450,9 @@ function RootDropdown({
         // textareas or non-input wrappers)
         const isCombobox =
             isSearchable && isTextInputTrigger && trigger.type !== 'textarea';
-        const triggerRole = triggerProps.role ?? (isCombobox ? 'combobox' : undefined);
+        const triggerRole =
+            triggerProps.role ??
+            (isCombobox ? 'combobox' : isMenubarItem ? 'menuitem' : undefined);
         // aria-expanded isn’t supported on textbox either, so a text input
         // only gets one once a role makes it something else
         const canExpand = !isTextInputTrigger || triggerRole != null;
@@ -1450,13 +1466,17 @@ function RootDropdown({
             'aria-haspopup': triggerProps['aria-haspopup'] ?? popupRole,
             id: resolvedTriggerId,
             role: triggerRole,
+            tabIndex: triggerProps.tabIndex ?? menubarTabIndex,
         });
     }
 
     if (label != null) {
         bodyLabelledBy = labelId;
         trigger = (
-            <label className="uktdropdown-label">
+            <label
+                className="uktdropdown-label"
+                role={isMenubarItem ? 'none' : undefined}
+            >
                 <div className="uktdropdown-label-text" id={labelId}>
                     {label}
                 </div>
@@ -1477,6 +1497,7 @@ function RootDropdown({
                     'is-searchable': isSearchable,
                 })}
                 onClick={onClick}
+                role={isMenubarItem ? 'none' : undefined}
                 onMouseDown={handleMouseDown}
                 onMouseEnter={handleDropdownMouseEnter}
                 onMouseLeave={handleDropdownMouseLeave}
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -297,24 +297,23 @@ function RootDropdown({
     const labelId = `${bodyId}-label`;
     const popupRole = isSearchable ? 'listbox' : hasItems ? 'menu' : 'dialog';
     const menubar = useContext(MenubarContext);
-    // A menubar owns menuitems, so a member's trigger takes role="menuitem"
-    // and the elements between the bar and it are neutralized — a generic div
-    // (or label) in between leaves the menubar owning no menuitems at all.
-    // Only a menu trigger qualifies: a searchable member is a combobox, which
-    // isn't a valid menubar child, so it keeps its own semantics.
-    const isMenubarItem = menubar != null && popupRole === 'menu';
+    // Every member’s wrapper elements are neutralized, so the bar owns the
+    // controls themselves rather than the generic divs (and label) around
+    // them — a generic in between leaves the menubar owning no menu items at
+    // all. This applies to any member, including a searchable one.
+    const isMenubarMember = menubar != null;
+    // Only a member whose popup is a menu joins the roving tabindex, and takes
+    // role=menuitem unless the trigger declares its own. A searchable member is
+    // a combobox, which is neither a valid menubar child nor part of the bar’s
+    // tab stop, so it keeps its own semantics.
+    const isMenubarItem = isMenubarMember && popupRole === 'menu';
     // The bar keeps a single tab stop (APG roving tabindex); the members that
-    // don't hold it stay reachable with ←/→.
+    // don’t hold it stay reachable with ←/→.
     //
     // Before registration both sides are null, so every member renders
-    // tabbable on the first render. That's deliberate: the first render is
-    // also the SSR render, and the roving tabindex is only navigable because
-    // JS is running (←/→ is the Menubar's own key handler). Without JS, all
-    // triggers tabbable degrades to plain buttons — every menu still
-    // reachable — whereas defaulting them to -1 would leave the bar with no
-    // keyboard entry point and no arrow navigation to compensate. On the
-    // client the window is a single render: members register in their effects
-    // during the mount commit, so the first paint already has one stop.
+    // tabbable on the first render to ensure the UI works without JS. Members
+    // proceed to register in their effects, so the bar settles to one stop as
+    // soon as those run.
     const menubarTabIndex = !isMenubarItem
         ? undefined
         : menubar.tabbableElement === dropdownElement
@@ -674,11 +673,9 @@ function RootDropdown({
             element: dropdownElement,
             focusTrigger,
             isDisabled: () => Boolean(disabled),
-            // Only a member rendered as a menuitem takes part in the bar's
-            // roving tabindex; a searchable member is a combobox with its own
-            // native tab stop, and letting it hold the bar's would leave every
-            // menuitem at -1
-            isMenuItem: () => isMenubarItem,
+            // Only a member whose popup is a menu takes part in the bar’s
+            // roving tabindex
+            isMenuPopup: () => isMenubarItem,
             isOpen: () => isOpenRef.current,
             // Gated as well as skipped by the menubar: open() is the one
             // opening path that doesn’t go through this component’s own
@@ -1490,12 +1487,12 @@ function RootDropdown({
         trigger = (
             <label
                 className="uktdropdown-label"
-                role={isMenubarItem ? 'none' : undefined}
+                role={isMenubarMember ? 'none' : undefined}
             >
                 <div
                     className="uktdropdown-label-text"
                     id={labelId}
-                    role={isMenubarItem ? 'none' : undefined}
+                    role={isMenubarMember ? 'none' : undefined}
                 >
                     {label}
                 </div>
@@ -1516,7 +1513,7 @@ function RootDropdown({
                     'is-searchable': isSearchable,
                 })}
                 onClick={onClick}
-                role={isMenubarItem ? 'none' : undefined}
+                role={isMenubarMember ? 'none' : undefined}
                 onMouseDown={handleMouseDown}
                 onMouseEnter={handleDropdownMouseEnter}
                 onMouseLeave={handleDropdownMouseLeave}
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -304,7 +304,17 @@ function RootDropdown({
     // isn't a valid menubar child, so it keeps its own semantics.
     const isMenubarItem = menubar != null && popupRole === 'menu';
     // The bar keeps a single tab stop (APG roving tabindex); the members that
-    // don't hold it stay reachable with ←/→
+    // don't hold it stay reachable with ←/→.
+    //
+    // Before registration both sides are null, so every member renders
+    // tabbable on the first render. That's deliberate: the first render is
+    // also the SSR render, and the roving tabindex is only navigable because
+    // JS is running (←/→ is the Menubar's own key handler). Without JS, all
+    // triggers tabbable degrades to plain buttons — every menu still
+    // reachable — whereas defaulting them to -1 would leave the bar with no
+    // keyboard entry point and no arrow navigation to compensate. On the
+    // client the window is a single render: members register in their effects
+    // during the mount commit, so the first paint already has one stop.
     const menubarTabIndex = !isMenubarItem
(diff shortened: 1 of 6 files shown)


packages/dropdown/src/Dropdown.tsx
```diff
@@ -304,6 +304,19 @@ function RootDropdown({
...
+    // and the elements between the bar and it are neutralized — a generic div
+    // (or label) in between leaves the menubar owning no menuitems at all.
+    // Only a menu trigger qualifies: a searchable member is a combobox, which
+    // isn't a valid menubar child, so it keeps its own semantics.
+    const isMenubarItem = menubar != null && popupRole === 'menu';
+    // The bar keeps a single tab stop (APG roving tabindex); the members that
+    // don't hold it stay reachable with ←/→
+    const menubarTabIndex = !isMenubarItem
+        ? undefined
+        : menubar.tabbableElement === dropdownElement
```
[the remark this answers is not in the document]
acusti: Correct on the mechanism — `null === null` on the first render, so every enabled member does render `tabIndex={0}` until registration lands. I'm keeping it, though, because I think the alternative is worse where it actually matters.

The first render is also the **SSR** render, and the roving tabindex is only navigable *because* JavaScript is running: ←/→ moving focus between triggers is `Menubar`'s own key handler. So the two candidate no-JS outcomes are:

- **all triggers tabbable** (today): every menu reachable by Tab. Degrades to plain buttons, which is what they are without JS.
- **all triggers at `-1`** (the suggestion): the menubar is unreachable by keyboard entirely, with no arrow navigation to compensate.

Trading "temporarily more tab stops than the pattern calls for" for "temporarily no keyboard entry point at all" isn't a trade I want to make on markup that ships to users before hydration.

On the client the window is one render — registration happens in the members' effects, in the same commit as mount, so the first paint after hydration already has the single stop. It's not observable to a user, and there's no interactive moment during which the extra stops are reachable.

The three genuinely stale-state versions of this — an unmounted holder, a disabled holder, and (new in 9e7db31) a member unmounted by an intermediate component without `Menubar` re-rendering — are the ones that would strand the bar, and those are all handled.

I've added a comment at the computation recording this reasoning so it doesn't read as an oversight next time. Happy to flip it if you'd rather the pattern hold literally from the first byte — it's a one-line change.

---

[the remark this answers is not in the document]
acusti: Following up on the two newest findings, both fixed in 3e61363 — replying here since their own threads are on outdated lines.

**Searchable member's generic wrapper.** Valid. Wrapper neutralization was gated on the same predicate that decides whether a trigger becomes a `menuitem`, so a searchable member kept its `div.uktdropdown` as a generic — the same `aria-required-children` failure, in a different shape. Membership and menuitem-ness are now separate: any member's wrappers are neutralized, while only a menu trigger takes `role="menuitem"` and joins the roving tabindex.

One correction to the suggested strategy, though — "keep the combobox out of the menubar's owned children" isn't achievable. Ownership follows the DOM; there's no way to opt an element out short of not putting it in the bar. And a `combobox` isn't allowed menubar content under *any* arrangement (`menubar` permits `group`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `separator`). So neutralizing removes the generic but doesn't make the member valid — that's a property of the combination, not the markup. The README now says a search field belongs outside the `Menubar`, since a container mixing menus with other controls is a `toolbar`. The test pins the wrappers and names the combobox as the known-invalid part rather than blessing it in the allow-list.

**Test only ran against a closed bar.** Also valid, and the more useful of the two. It now runs with a menu open, which is where the interesting ownership happens: the body mounts inside the neutralized wrapper, so the bar owns it directly.

Worth noting that state turns out to be *correct* — it mirrors the canonical APG menubar, where the menuitem and its menu sit inside the same presentational wrapper. So the open-bar assertion documents the intended shape rather than catching a bug.

You were also right that `expect.any(String)` was too loose — it could only catch null-role generics. It's now an explicit allow-list (`menuitem`, `menu`, `none`), so a mis-roled owned element fails too.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -297,6 +297,36 @@ function RootDropdown({
     const labelId = `${bodyId}-label`;
     const popupRole = isSearchable ? 'listbox' : hasItems ? 'menu' : 'dialog';
     const menubar = useContext(MenubarContext);
+    // Every member’s wrapper elements are neutralized, so the bar owns the
+    // controls themselves rather than the generic divs (and label) around
+    // them — a generic in between leaves the menubar owning no menu items at
+    // all. This applies to any member, including a searchable one: its
+    // combobox isn’t valid menubar content either way, and leaving a generic
+    // wrapper around it only adds a second problem on top of the first (see
+    // the README — a search field belongs outside the bar).
```
acusti: the searchable combobox stuff is covered in the next comment one-line down from here
```suggestion
    // all. This applies to any member, including a searchable one.
```

packages/dropdown/src/Dropdown.tsx
```diff
@@ -297,6 +297,36 @@ function RootDropdown({
...
+    //
+    // Before registration both sides are null, so every member renders
+    // tabbable on the first render. That’s deliberate: the first render is
+    // also the SSR render, and the roving tabindex is only navigable because
+    // JS is running (←/→ is the Menubar’s own key handler). Without JS, all
+    // triggers tabbable degrades to plain buttons — every menu still
+    // reachable — whereas defaulting them to -1 would leave the bar with no
+    // keyboard entry point and no arrow navigation to compensate. On the
+    // client the window is a single render: members register in their effects
+    // during the mount commit, so the first paint already has one stop.
```
```suggestion
    // tabbable on the first render to ensure the UI works without JS. Members
    // proceed to register in their effects during the mount commit, so the first
    // paint already has one stop.
```

packages/dropdown/src/Dropdown.tsx
```diff
@@ -651,6 +681,11 @@ function RootDropdown({
             element: dropdownElement,
             focusTrigger,
             isDisabled: () => Boolean(disabled),
+            // Only a member whose popup is a menu takes part in the bar’s
+            // roving tabindex; a searchable member is a combobox with its own
+            // native tab stop, and letting it hold the bar’s would leave every
+            // menu item at -1
```
```suggestion
            // roving tabindex
```

packages/dropdown/src/Menubar.tsx
```diff
@@ -54,11 +57,23 @@ const findNextEnabledMember = (
     return null;
 };
 
+// Only an enabled member whose popup is a menu can hold the bar’s tab stop. A
+// searchable member is a combobox with its own native tab stop and never reads
+// the roving value, so letting it take ownership would leave every menu item at
+// -1 and the bar without a menu entry point.
```
acusti: i don’t think we need to cover the searchable member combobox thing again here
```suggestion
// Only an enabled member whose popup is a menu can hold the bar’s tab stop
```

packages/dropdown/src/Menubar.tsx
```diff
@@ -69,6 +84,26 @@ export default function Menubar({ children, className, style }: MenubarProps) {
...
+    // falls to the first enabled member in document order. Without this an
+    // unmounted or disabled owner would leave the bar with no tab stop at all.
+    //
+    // Deliberately runs on every render rather than on a dependency list. An
+    // unmounting member changes no state here — it only removes itself from
+    // membersRef during its own cleanup — so there is nothing to depend on
+    // that would catch it. The setState the lint rule warns about can’t chain:
+    // it only fires when the current holder is missing or disabled, and the
+    // value it sets makes the guard above return on the next run.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
acusti: did you run the build and inspect it to verify that this still gets compiled by react compiler with this suppression?

packages/dropdown/src/Menubar.tsx
```diff
@@ -104,12 +139,37 @@ export default function Menubar({ children, className, style }: MenubarProps) {
...
+                    // since React runs every cleanup and setup before the
+                    // batched re-render.
+                    reconcile();
+                    // Deliberately not released here. Members re-register on
+                    // every render, and React runs every cleanup before any
+                    // setup — so releasing would hand the tab stop to whichever
+                    // member re-registers first (always the first in the bar)
+                    // rather than back to the one that held it. A genuinely
+                    // unmounted holder is picked up by the effect below, which
+                    // no longer finds it among the members.
```
acusti: can we simplify these comment blocks? the ratio of code-to-comment is out of control and it makes my eyes glaze over

packages/dropdown/src/Menubar.tsx
acusti: No, I hadn't — and it wasn't compiled. Good catch. Fixed in d422d05.

**What the build actually said.** `Menubar` came out of `dist/Dropdown.js` as a plain function, while `RootDropdown` in the same file starts `const $ = c(181);`. So the compiler was running on the file and skipping this component. Two separate bailouts, both reported by the logger already configured in `vite.config.base.js`:

1. **`PreserveManualMemo`** — *"The inferred dependency was `reconcile`, but the source dependencies were [tabbableElement]."* The `contextValue` `useMemo` closes over the `useReducer` dispatch without listing it. React guarantees dispatch stability; the compiler doesn't assume it, so it bailed on the whole component.
2. **The suppression itself** — *"React Compiler has skipped optimizing this component because one or more React ESLint rules were disabled. Found suppression `eslint-disable-next-line react-hooks/exhaustive-deps`."* Exactly what you suspected.

Worth noting they masked each other: removing the suppression alone left it uncompiled (bailout 1 still applied), which is how I could have "checked" and drawn the wrong conclusion.

**The fix removes the suppression rather than justifying it.** The rule was complaining about `setState` in a dep-less effect, and the effect was dep-less because I'd concluded a member leaving "changes no state here". That was wrong — the cleanup already calls `reconcile()`, so there *is* a signal:

```js
const [reconcileCount, reconcile] = useReducer((count: number) => count + 1, 0);
…
}, [reconcileCount, tabbableElement]);
```

Same coverage — mount, owner change, and any member registering or unregistering — with a real dependency list, no suppression, and no lint complaint. Plus `[reconcile, tabbableElement]` on the `useMemo` for bailout 1.

**Verified after:** zero `React Compiler logger` events across the package build, and `function Menubar(t0) { const $ = c(22); … }` in `dist`. 317 tests pass, lint clean without the suppression, tsc clean.

---

packages/dropdown/src/Menubar.tsx
acusti: Done in d422d05 — `Menubar.tsx` is down from 77 comment lines to 61, with the reductions concentrated in the two blocks you're pointing at.

`registerMember` was the worst offender: 5 lines to 2, and its cleanup 14 lines to 6.

```js
            registerMember(member: MenubarMember) {
                membersRef.current.add(member);
                // Claim the tab stop when it’s going spare; members register
                // in document order, so the first enabled one gets it
                setTabbableElement((current) =>
                    current == null && canHoldTabStop(member) ? member.element : current,
                );
                return () => {
                    membersRef.current.delete(member);
                    // say the set has settled, so the effect above can pick
                    // up a tab stop stranded on an unmounted member
                    reconcile();
                    // the tab stop is deliberately not released here: members
                    // re-register on every render, and React runs every cleanup
                    // before any setup, so releasing would hand it to whichever
                    // member re-registers first rather than back to the holder
                };
            },
```

What went was the derivation — *why* a member's effect doesn't re-render `Menubar`, *how* an intermediate component can unmount one, *when* React runs cleanups relative to setups. What stayed is the part that's load-bearing if you're editing the line: don't release the tab stop here, and why releasing would be wrong.

The reconciliation effect above shrank too, from 12 lines to 3, mostly because the compiler fix let me delete the paragraph justifying the suppression — the dependency list now says what the prose was arguing for.

---

packages/dropdown/src/Dropdown.tsx
acusti: In as written (d422d05). You're right that it's the next comment's job — this one is about wrapper neutralization, and "including a searchable one" is all it needs to say about the exemption.

That combobox explanation was repeated in four places across this PR; three of them are gone now, leaving the one on `isMenubarItem` that actually decides it.

---

packages/dropdown/src/Dropdown.tsx
acusti: In as written (d422d05), 9 lines to 4. "to ensure the UI works without JS" carries the whole argument I'd spelled out over five lines — the reader doesn't need the enumeration of what the -1 alternative would cost to accept it.

---

packages/dropdown/src/Dropdown.tsx
acusti: In as written (d422d05) — 4 lines to 2. `isMenuPopup` is a one-line predicate; the consequences of getting it wrong belong at the call site, not on the definition.

---

packages/dropdown/src/Menubar.tsx
acusti: In as written (d422d05) — 4 lines to 1. Agreed it was the fourth telling; `isMenubarItem` is where that decision lives, and the rest were just restating it.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -297,6 +297,28 @@ function RootDropdown({
...
+    // a combobox, which is neither a valid menubar child nor part of the bar’s
+    // tab stop, so it keeps its own semantics.
+    const isMenubarItem = isMenubarMember && popupRole === 'menu';
+    // The bar keeps a single tab stop (APG roving tabindex); the members that
+    // don’t hold it stay reachable with ←/→.
+    //
+    // Before registration both sides are null, so every member renders
+    // tabbable on the first render to ensure the UI works without JS. Members
+    // proceed to register in their effects during the mount commit, so the
+    // first paint already has one stop.
```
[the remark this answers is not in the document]
acusti: Correct — `useEffect` is passive, so it runs after paint, and "the first paint already has one stop" claimed the opposite. Fixed in 53ed302:

```js
    // Before registration both sides are null, so every member renders
    // tabbable on the first render to ensure the UI works without JS. Members
    // proceed to register in their effects, so the bar settles to one stop as
    // soon as those run.
```

The behavior is unchanged and still fine — the extra tab stops exist for one paint, with no interactive moment in between — but the comment now says when they go away rather than asserting they were never there.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -1450,14 +1478,22 @@ function RootDropdown({
             'aria-haspopup': triggerProps['aria-haspopup'] ?? popupRole,
             id: resolvedTriggerId,
             role: triggerRole,
+            tabIndex: triggerProps.tabIndex ?? menubarTabIndex,
```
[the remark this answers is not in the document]
acusti: Accurate about the mechanism, but this is deliberate and staying as-is.

`triggerProps.tabIndex ?? menubarTabIndex` is the same your-values-win rule every other fill-in on that `cloneElement` follows — `aria-controls`, `aria-expanded`, `aria-haspopup`, `role`, `id`. Enforcing the roving value when `isMenubarItem` would make `tabIndex` the single prop the component overrides you on, which is a worse inconsistency than the failure mode it prevents.

It's also the documented posture, in the accessibility section on `main`:

> The ARIA it adds is a best effort from what it can control: it annotates the DOM it renders and the shapes it recognizes, but never overwrites a value you pass. So when you provide a completely custom trigger, it is possible to end up with ARIA that doesn't hold together. The safest thing if you go full custom is to take over the ARIA as well.

Setting `tabIndex={0}` on two custom triggers in a bar is a consumer choosing that, the same way a `<div>` trigger can end up carrying an `aria-expanded` its role doesn't support. The component's job ends at not doing it *to* you.

---

PR #417, opened 2026-08-10
Add @acusti/vite-plugin-react-compiler

acusti:
Adds a new package, `@acusti/vite-plugin-react-compiler`: a Vite plugin that runs React Compiler via [oxc-transform-react](https://www.npmjs.com/package/oxc-transform-react) (the native Node bindings for the oxc project’s Rust port of the compiler), as a drop-in alternative to the Babel-based React Compiler build path. The design was validated in a production Vite 8 + React Router app: full memoization parity with babel-plugin-react-compiler@1.0.0 across 987 source files with zero regressions, a ~16x faster compiler pass, and a ~2.5x faster full production build (53s → 21s).

## Implementation

- `transform`-hook plugin with `enforce: 'pre'`, so the compiler sees original component code before any refresh/JSX transforms — the same position the Babel pass occupies via vite-plugin-babel today
- Filters modules with Vite’s `createFilter`; defaults `include: /\.[jt]sx?$/`, `exclude: /[/\\]node_modules[/\\]/` (matching @rolldown/plugin-babel’s default), both user-configurable; query-suffixed ids are normalized to the underlying file path (like @vitejs/plugin-react)
- Calls the async `transform` of oxc-transform-react with `jsx: 'preserve'` (Vite’s own oxc transform stays in charge of JSX handling downstream), `sourcemap: true`, and the user’s `reactCompiler` options passed through verbatim (same option names as babel-plugin-react-compiler)
- Fails the build via `this.error(...)` when the result is `fatal` (parse failure or rejected options), like a Babel syntax error would — including oxc’s code frames and the error’s source position (converted from oxc’s utf-8 byte offset to a string index)
- Default-on transform memoization keyed by file path + sha1 content hash, storing the in-flight promise so overlapping transforms (concurrent client + SSR environments) share a single compiler run
- `oxc-transform-react` is a regular dependency pinned exactly (currently 0.144.0): the bindings track React Compiler main on a fast-moving 0.x release train of near-100% bug-fix releases, so the plugin pins the newest version that passes this repo’s test suite and ships updates as regular plugin releases (escape hatch: package-manager `overrides`/`resolutions`)
- The `vite` peer range is `>=8.0.0`, intentionally: like the sibling vite-plugin-svg-react, this package targets the oxc/rolldown-native pipeline that is the whole point of the plugin

## Structure

Mirrors `packages/vite-plugin-svg-react`: same package.json layout and publishing config, shared `vite.config.base.js` build with `target: 'node20'`, matching tsconfig pair, README style, and a changeset for the initial 0.1.0 minor release. Registered in the root tsconfig references and root README package table.

## Tests

Two suites, ten tests total, following the repo’s existing direct-hook-invocation approach.

`index.test.ts` (plugin behavior): memoized output via `react/compiler-runtime` with JSX preserved and TypeScript stripped, query-suffix id normalization, include/exclude filtering (including that `node_modules` lookalike segments still compile), `reactCompiler` option passthrough (`target: '18'` → `react-compiler-runtime`; `compilationMode: 'annotation'` skips unannotated components), sourcemap presence, fatal-error handling via `this.error` with the correct position (multi-byte-safe), and memoization (repeat, concurrent, content-change, and opt-out cases).

`upstream.test.ts` (pinned-bindings tracking): currently four enabled pins and no tracked divergences — improvements over babel-plugin-react-compiler@1.0.0 from tracking React Compiler main (computed object keys and try/catch value blocks both compile; upstream confirmed try/catch conforms to the unpublished Babel compiler and dropped the bailout restoration — oxc-project/oxc#25343 closed as not planned, #25409 closed unmerged) and Babel-parity fixes as they land (react-hooks lint suppressions opt functions out since 0.144.0 — that tracking test flipped red on the bump exactly as designed and was promoted to a pin). Future known divergences get `it.fails` tests that flip red on the bindings bump that changes them.

## Validation

`bun run build`, `bun run test`, `bun run lint`, `bun run tsc`, and `bun run format:check` are all green.

One note for the release: the repo is currently in changesets pre-release (beta) mode, so `changeset version` would give the new package a `0.1.0-beta.x` version until pre mode exits.

https://claude.ai/code/session_01VvbKw4yfxAnTYtdWZgwvmR


packages/vite-plugin-react-compiler/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+                // leave JSX for vite’s own pipeline (refresh, dev runtime)
+                jsx: 'preserve',
+                reactCompiler: options.reactCompiler ?? {},
+                sourcemap: true,
+            });
+
+            // fatal means a parse failure or rejected options: no code was
+            // emitted, so break the build like a Babel syntax error would
+            if (transformed.fatal) {
+                this.error(transformed.errors.map((error) => error.message).join('\n'));
```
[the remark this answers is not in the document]
acusti: Fixed in 2636dda. Fatal errors now include oxc’s code frames (which carry file:line:column, e.g. `,-[/src/broken.tsx:2:7]`) in the message passed to `this.error(...)`, and the first error’s label position is forwarded as the `pos` argument so vite can attach the location. The parse-error test asserts both the code-frame content and the forwarded position.

---

packages/vite-plugin-react-compiler/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+            if (!filter(id)) return null;
+
+            let hash = '';
+            if (isMemoizing) {
+                hash = createHash('sha1').update(code).digest('hex');
+                const cached = cache.get(id);
+                if (cached?.hash === hash) return cached.result;
+            }
+
+            const transformed = await transform(id, code, {
```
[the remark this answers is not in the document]
acusti: Fixed in 2636dda. The cache now stores the in-flight transform promise instead of the completed result, so overlapping transforms of identical content share a single compiler run. Rejected transforms are evicted from the cache so the next request retries instead of replaying a cached failure. Covered by new assertions in the memoization test (parallel un-awaited calls resolve to the same result object) and the parse-error test (an identical retry runs the compiler again).

---

packages/vite-plugin-react-compiler/src/index.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    it('returns a sourcemap', async () => {
+        const transform = createTransform();
+        const result = await transform(COMPONENT, '/src/Greeting.tsx');
+        expect(result?.map?.version).toBe(3);
+        expect(result?.map?.mappings.length).toBeGreaterThan(0);
+    });
+
+    it('errors on fatal input (parse errors)', async () => {
+        const transform = createTransform();
+        await expect(transform('const = ;', '/src/broken.tsx')).rejects.toThrow();
```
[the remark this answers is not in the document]
acusti: Fixed in 2636dda. The test context now records every `this.error(message, pos)` invocation, and the parse-error test asserts exactly one call with the expected message content (including the code frame’s file:line:column) and forwarded position — pinning the plugin’s own `fatal` detection path rather than any upstream rejection.

---

packages/vite-plugin-react-compiler/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+            // fatal means a parse failure or rejected options: no code was
+            // emitted, so break the build like a Babel syntax error would
+            if (transformed.fatal) {
+                this.error(transformed.errors.map((error) => error.message).join('\n'));
+            }
+
+            const result = { code: transformed.code, map: transformed.map };
+            if (isMemoizing) cache.set(id, { hash, result });
+            return result;
```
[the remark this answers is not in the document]
acusti: Fixed in f120ced. The transform now splits the query off the id and uses the bare file path for the include/exclude filter, the cache key, and the filename passed to oxc-transform-react — mirroring @vitejs/plugin-react’s id handling. Covered by a new test asserting a `?v=`-suffixed id still compiles, shares the query-less id’s cache entry, and produces a sourcemap whose `sources` contain only the clean path.

Two notes on the premise: `?t=` and `?import` specifically never reach transform hooks (vite strips them while resolving the request), and previously a query-suffixed id was rejected by the default extension-anchored include filter rather than cached separately — so the practical effect of the change is that query-carrying ids (e.g. sub-request queries from other plugins) now get compiled and cached per-file instead of silently skipping the compiler pass, which is the better behavior anyway.

---

packages/vite-plugin-react-compiler/src/index.ts
```diff
@@ -0,0 +1,113 @@
...
+                if (transformed.fatal) {
+                    this.error(
+                        transformed.errors
+                            .map((error) =>
+                                error.codeframe
+                                    ? `${error.message}\n${error.codeframe}`
+                                    : error.message,
+                            )
+                            .join('\n'),
+                        transformed.errors[0]?.labels[0]?.start,
```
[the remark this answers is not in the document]
acusti: Fixed in 648190d. Confirmed empirically against the installed bindings: for a source with `é`/`💜` before the failing token, oxc reports `start: 30` where the JS string index is 27 — so the label offsets are indeed utf-8 byte offsets. The plugin now decodes the byte prefix of the code (`Buffer.from(code, 'utf8').subarray(0, byteStart).toString('utf8').length`) to convert to a string index before forwarding to `this.error`, and the parse-error test now places multi-byte characters before the error and asserts the forwarded position equals the string index, so it can no longer pass by ascii coincidence.

---

packages/vite-plugin-react-compiler/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+[@rolldown/plugin-babel]:
+    https://www.npmjs.com/package/@rolldown/plugin-babel
+
+## Usage
+
+```
+npm install --save-dev @acusti/vite-plugin-react-compiler oxc-transform-react
+# or
+yarn add --dev @acusti/vite-plugin-react-compiler oxc-transform-react
```
acusti: let’s add pnpm / bun examples

acusti: Done in 659ee61 — the install block now shows npm, yarn, pnpm, and bun variants.

---

packages/vite-plugin-react-compiler/README.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+Note that [oxc-transform-react][] is a peer dependency, not a dependency:
+it is on a fast-moving 0.x release train tracking React Compiler main, so
+you should pin your own exact version of it to make compiler behavior
+changes arrive deliberately, on your schedule.
```
acusti: i’m second-guessing this. i agree that it’s a fast-moving 0.x release train, but that feels like all the more reason we should commit to keeping this plugin up-to-date and constantly catching the newest release and testing it and publishing new versions. right now, new versions are likely to be close to if not 100% bug fixes, due to this being a port of an existing widely-used tool, as opposed to other kinds of packages, which might use pre-1.0 versions to release experimental new features or breaking changes.

packages/vite-plugin-react-compiler/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+Like [@acusti/vite-plugin-svg-react][], this plugin requires Vite ≥ 8, on
+purpose: its reason for existing is completing the all-native oxc/rolldown
+pipeline that Vite 8 introduced. On Vite < 8, the Babel-based
+`reactCompilerPreset` path is no slower than the rest of the pipeline.
```
acusti:
```suggestion
pipeline that Vite 8 introduced.
```

packages/vite-plugin-react-compiler/README.md
acusti: Agreed and done in d67bffe: oxc-transform-react is now a regular dependency pinned exactly to 0.142.0 (no more peerDependency), so each plugin release ships the newest bindings version that passes this repo’s test suite, and consumers just install the plugin. Dependabot’s root npm config already covers the workspace, so new bindings releases will arrive as bump PRs that run the full suite before publishing.

The README paragraph now makes your argument explicitly (fast-moving 0.x train + releases that are close to 100% bug fixes → the plugin owns the version), keeps an escape hatch via package-manager `overrides`/`resolutions`, and the install commands drop the second package. One mechanical note: an exact pin (rather than `^0.142.0`) is deliberate — oxc bumps the minor on every release, so for 0.x a caret would never float anyway, and the exact pin makes each upgrade an explicit, tested bump.

---

packages/vite-plugin-react-compiler/README.md
acusti: Applied in d67bffe — the paragraph now ends at “pipeline that Vite 8 introduced.”

---

packages/vite-plugin-react-compiler/src/index.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+        // annotation mode skips components without a 'use memo' directive
+        const annotation = createTransformer({
+            reactCompiler: { compilationMode: 'annotation' },
+        });
+        const annotationResult = await annotation.transformCode(
+            COMPONENT,
+            '/src/Greeting.tsx',
+        );
+        expect(annotationResult?.code).not.toContain('compiler-runtime');
```
acusti: let’s standardize on a default method for verifying the compiler ran (or `.not.` ran), either `.toContain('react/compiler-runtime')` or `.toContain('compiler-runtime')`.

i think i’d lean towards the version that includes the `react/` suffix in order to make the distinction between the default behavior and the `target: '18'` behavior, but i’d be fine with either if there’s a reason to go the other way

acusti: Standardized on `'react/compiler-runtime'` in 2d47571 — agreed it’s the better default since it distinguishes the default target from `target: '18'`. This annotation-mode assertion was the only bare-`'compiler-runtime'` use; it’s now `.not.toContain('react/compiler-runtime')`, which is also the precise regression check under default options (an unexpectedly-compiled component would import exactly that specifier). The `target: '18'` test keeps `'react-compiler-runtime'` as the one deliberate exception, since asserting the 18 runtime is that test’s whole point.

---

.changeset/vite-plugin-react-compiler-initial-release.md
```diff
@@ -0,0 +1,17 @@
...
+the Babel-based build path. The plugin runs the compiler pass in a
+`transform` hook with `enforce: 'pre'` and `jsx: 'preserve'`, so Vite’s own
+oxc pipeline stays in charge of JSX/refresh/TypeScript handling downstream,
+and memoizes transform results by module id + content hash so
+multi-environment builds only compile each file once. Fatal compiler errors
+(parse failures, rejected options) fail the build like a Babel syntax error
+would. The plugin depends on an exact, tested version of
+oxc-transform-react — the bindings track React Compiler main on a
+fast-moving 0.x release train, so updates ship as regular plugin releases
+after passing this repo’s test suite.
```
acusti, line 17: I like this thought, but it makes me wonder if we should add an additional test suite that goes through some of the possible issues/things that might not be handled that we hope will be handled in subsequent versions of `oxc-transform-react`. we could have all of the test cases that still fail be commented-out/skipped and bring them back every time a new oxc-transorm-react version is released and, if the issue is fixed, keep it enabled and it becames a new pin/regression test.

for example, one of the cool things about `oxc-transform-react` is that it already improves over the babel-based react-compiler implementation. when we ran it in outlyne, we saw that it actually compiles a strict superset (7 extra functions in 6 files, because the Rust port tracks compiler main, which supports `try/catch` value blocks and computed object keys that `babel-plugin-react-compiler@1.0.0` bails on).

another example: https://github.com/oxc-project/oxc/issues/25392. we could have a test to see if a component with `// eslint-disable-next-line react-hooks/exhaustive-deps` (or the oxlint equivalent is skipped or if it’s still processed, e.g.:
```
// eslint-disable-next-line react-hooks/exhaustive-deps
export function Counter({ step }) {
  const [count, setCount] = useState(0);
  const ref = useRef(step);

  useEffect(() => {
    setCount((c) => c + ref.current);
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);

  return <div>{count}</div>;
}
```
and once that’s fixed, we can enable that test to pin it. normally i strictly avoid testing the internals of dependencies (a big point of the dep is to let it handle its own verifications/tests), but if we are offering this kind of assurance/value, might make sense to break that rule.

acusti: Built in 35444c5 as `src/upstream.test.ts`, with one refinement on the mechanics: instead of commented-out/skipped cases that need manual re-enabling on each release, known gaps are written as vitest `it.fails` tests asserting the *desired* behavior. Today the assertion fails, so the suite is green; when a bindings bump fixes the gap, the assertion starts passing, `it.fails` turns red on that bump PR, and the test gets promoted to a plain `it` as a new pinned regression test. That makes the “re-check on every release” step automatic — the bump PR itself tells you when to flip a test on.

Contents, verified against the pinned 0.142.0 before writing the assertions:
- **Enabled pins (improvements over babel-plugin-react-compiler@1.0.0):** try/catch value blocks and computed object keys both compile (both confirmed compiling via `react/compiler-runtime`), so a bindings regression fails the bump PR.
- **Tracked gap (`it.fails`):** your exact `Counter` example with `// eslint-disable-next-line react-hooks/exhaustive-deps` — confirmed the suppression does *not* opt the function out in 0.142.0 (it still compiles), matching oxc-project/oxc#25392. The test asserts the Babel-plugin behavior (skipped) and will flip red when upstream wires suppressions up.

Also added a sentence to the README’s caveats section so the upstream-tracking suite is part of the published “each release is tested” story. The suite header comment documents the convention (including that it deliberately breaks the don’t-test-your-deps rule and why).

---

packages/vite-plugin-react-compiler/src/index.ts
```diff
@@ -0,0 +1,121 @@
...
+};
+
+type CachedTransform = {
+    hash: string;
+    result: Promise<TransformOutput>;
+};
+
+type TransformOutput = { code: string; map?: SourceMap };
+
+const defaultExclude = /node_modules/;
```
acusti: should we try to match only exactly `node_modules` (as opposed to also matching `not_actually_node_modules`)? e.g. 
```suggestion
const defaultExclude = /[\b/]node_modules[\b/]/;
```
i just wanna match the existing vite react compiler babel plugin behavior, so we go with whatever matches that.

packages/vite-plugin-react-compiler/src/index.ts
```diff
@@ -0,0 +1,121 @@
...
+                });
+
+                // fatal means a parse failure or rejected options: no code
+                // was emitted, so break the build like a Babel syntax error
+                // would, forwarding the first error’s position so vite can
+                // report the location alongside oxc’s own code frames
+                if (transformed.fatal) {
+                    // oxc label offsets are utf-8 byte offsets, but rollup’s
+                    // pos argument indexes the (utf-16) code string
+                    const byteStart = transformed.errors[0]?.labels[0]?.start;
```
acusti, line 86: ideally this would’ve been red/greened and we would have a test pinning it

packages/vite-plugin-react-compiler/package.json
```diff
@@ -0,0 +1,55 @@
...
+        "build": "vite build && tsc -p tsconfig.build.json",
+        "tsc": "tsc --noEmit"
+    },
+    "dependencies": {
+        "oxc-transform-react": "0.142.0"
+    },
+    "devDependencies": {
+        "@types/node": "^26.1.0",
+        "typescript": "^7",
+        "vite": "^8.0.0-0",
```
acusti: i think v8 stable has been around long enough that we don’t need to include the pre-released
```suggestion
        "vite": "^8",
```

packages/vite-plugin-react-compiler/src/index.ts
acusti: Fixed in ef52a17, matched to the upstream defaults. I checked both Babel-path references in node_modules: @rolldown/plugin-babel (what this repo’s compiler pass runs through) uses `DEFAULT_EXCLUDE = [/[/\\]node_modules[/\\]|^\0rolldown\/runtime\.js$/]`, and @vitejs/plugin-react uses `/\/node_modules\//`. So the default is now `/[/\\]node_modules[/\\]/` — the plugin-babel form, which is the vitejs form plus Windows backslashes (the `^\0rolldown/runtime.js$` half is a rolldown-internal virtual module that doesn’t apply here). The filter test now pins that `/src/not_actually_node_modules/Greeting.tsx` still compiles.

One note on the suggested regex: inside a character class `\b` means backspace (`\x08`), not a word boundary, so `[\b/]` would have effectively required slashes-or-backspace — the separator-bounded form is what both upstreams use anyway.

---

packages/vite-plugin-react-compiler/src/index.ts
acusti: There is a test pinning it — the parse-error test places multi-byte chars (`é`/`💜`) before the failing token and asserts `errorCalls[0].pos` equals the JS string index (`broken.indexOf('= ;')` → 27), which the raw byte offset (30) fails. Fair point on sequencing though: the fix and that test landed together in 648190d, so CI never showed the red state (the red was demonstrated in a local probe script against the bindings, not in the suite). I just re-verified the pin is load-bearing by reverting the conversion locally: the test fails with `expected 30 to be 27`, and goes green with the conversion restored — so a regression here can’t land silently.

---

packages/vite-plugin-react-compiler/package.json
acusti: Applied in 52b9f0d — the vite devDependency is now `^8` (the `>=8.0.0` peer range is unchanged).

---

.changeset/vite-plugin-react-compiler-initial-release.md
acusti: love it, i’ve never used `it.fails` before, but it’s perfect. one thing: seems like maybe the `value blocks within a try/catch bailout` fix isn’t actually a fix? i came across this issue about oxlint and that same bailout not being called out by the linter: https://github.com/oxc-project/oxc/issues/25343
and this PR, which closes it by restoring the bailout behavior: https://github.com/oxc-project/oxc/pull/25409

acusti: Good catch — confirmed against oxc-project/oxc#25409 (it restores the Babel-matching bailout to close #25343, and doesn’t touch computed object keys). Recategorized in 86fee6f: the try/catch case is now an `it.fails` tracked divergence asserting the bailout (`.not.toContain('react/compiler-runtime')`), with the issue/PR linked in the test comment — so the bindings bump that picks up the restoration flips it red for promotion, exactly like the suppressions case, instead of failing the bump PR as a phantom “improvement regression.” Computed object keys remain the enabled improvement pin, and the README caveats sentence now describes try/catch as a watched divergence rather than an improvement. Suite is now 2 enabled pins + 2 tracked divergences (`8 passed | 2 expected fail`).

---

packages/vite-plugin-react-compiler/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "exports": {
+        ".": "./dist/index.js"
+    },
+    "scripts": {
+        "test": "vitest",
+        "build": "vite build && tsc -p tsconfig.build.json",
+        "tsc": "tsc --noEmit"
+    },
+    "dependencies": {
+        "oxc-transform-react": "0.144.0"
```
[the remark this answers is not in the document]
acusti: Fixed in c3df769 — the README caveats sentence now describes suppressions as a pinned Babel-parity fix (“react-hooks lint suppressions opt functions out of compilation since 0.144.0”) rather than a watched divergence, leaving try/catch as the remaining tracked one.

---

PR #418, opened 2026-08-10
Read props directly instead of mirroring them into refs

acusti:
Follow-up to the `disabledRef` thread on #411 — the standalone pass at the rest of the prop-mirror block. Based on #416 (the a11y stack tip), since it touches the same code; nothing here is a11y-related.

## What went

Six mirrors, all read only from per-render closures (`handleSubmitItem`, the pointer/key handlers, `handleInput`), which are rebuilt whenever the props they read change:

`allowCreateRef` · `allowEmptyRef` · `hasItemsRef` · `keepOpenOnSubmitRef` · `onSubmitItemRef` · `valueRef`

## How it was verified

Not by test-passing, which turns out to prove nothing here. **The root `vitest.config.js` has no React plugin, so the suite exercises uncompiled source while the shipped build runs through `reactCompilerPreset`.** Under the compiler, "inline arrow, therefore rebuilt every render" isn't automatically true — handlers get memoized.

So I checked the built output instead. Each inlined prop appears in the memo guard of the handler that reads it; `handleSubmitItem`, which reads six of them, is gated on all six:

```js
if ($[64] !== allowCreate || $[65] !== allowEmpty || $[66] !== closeDropdown ||
    $[67] !== dropdownElement || $[68] !== handleActiveItem || $[69] !== hasItems ||
    $[70] !== keepOpenOnSubmit || $[71] !== onSubmitItem || $[72] !== popupRole ||
    $[73] !== valueIdentity) {
```

The compiler tracks each as a dependency, so the closure is rebuilt when it changes and the read can't go stale. (This also retroactively confirms the `disabledRef` removal in #411 against the real build — `handleRef` is memoized on `[closeDropdown, disabled, handleActiveItem]`.)

## What stayed, and why

Four mirrors earn their keep, for two distinct reasons — both now documented in place:

- **`isOpenRef` / `isOpeningRef`** are read by the document-level listeners attached in `handleRef`. Reading the state directly would make those closures depend on it, so every open and close would tear the listeners down and re-attach them. The mirror keeps the read live without the churn.
- **`onOpenRef` / `onCloseRef`** are invoked from an effect keyed on `[isOpen]` alone — that narrow dep list is what makes it fire on open/close *transitions* rather than whenever a consumer passes a fresh inline callback. Reading them directly would mean listing them as deps and firing on identity changes.

The non-mirror refs (timers, pointer tracking, `enteredCharactersRef`, etc.) are untouched — they hold mutable state that shouldn't trigger renders, which is what `useRef` is for.

## Testing

No new tests: this is a pure refactor with no observable behavior change, and the existing 156 pass unchanged. No changeset for the same reason.

`bun run test`, `lint`, `tsc`, `format`, and a compiled `build` all clean.

---

<a href="https://cubic.dev/pr/acusti/uikit/pull/418?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/dropdown/src/Dropdown.tsx
+++ b/packages/dropdown/src/Dropdown.tsx
@@ -342,40 +342,22 @@ function RootDropdown({
     const wasInSafeAreaRef = useRef<boolean>(false);
     const hoverCloseTimerRef = useRef<null | TimeoutID>(null);
 
-    const allowCreateRef = useRef(allowCreate);
-    const allowEmptyRef = useRef(allowEmpty);
-    const hasItemsRef = useRef(hasItems);
+    // The four values that still need mirroring, for two distinct reasons:
+    // 1. isOpen/isOpening are read by the document-level listeners attached in
+    //    handleRef
+    // 2. onOpen/onClose are called from an effect keyed on [isOpen] alone, which
+    //    is what makes it fire only on open/close transitions
     const isOpenRef = useRef(isOpen);
     const isOpeningRef = useRef(isOpening);
-    const keepOpenOnSubmitRef = useRef(keepOpenOnSubmit);
     const onCloseRef = useRef(onClose);
     const onOpenRef = useRef(onOpen);
-    const onSubmitItemRef = useRef(onSubmitItem);
-    const valueRef = useRef(valueIdentity);
 
     useEffect(() => {
-        allowCreateRef.current = allowCreate;
-        allowEmptyRef.current = allowEmpty;
-        hasItemsRef.current = hasItems;
         isOpenRef.current = isOpen;
         isOpeningRef.current = isOpening;
-        keepOpenOnSubmitRef.current = keepOpenOnSubmit;
         onCloseRef.current = onClose;
         onOpenRef.current = onOpen;
-        onSubmitItemRef.current = onSubmitItem;
-        valueRef.current = valueIdentity;
-    }, [
-        allowCreate,
-        allowEmpty,
-        hasItems,
-        isOpen,
-        isOpening,
-        keepOpenOnSubmit,
-        onClose,
-        onOpen,
-        onSubmitItem,
-        valueIdentity,
-    ]);
+    }, [isOpen, isOpening, onClose, onOpen]);
 
     // A consumer filtering or async-loading the body can drop or shift the
     // highlighted item without us seeing the change. Deliberately dep-less:
@@ -563,7 +545,7 @@ function RootDropdown({
     const trackSafeArea = (event: ReactMouseEvent<HTMLElement>) => {
         // Safe-area intent only applies to an open menu with items; skip the
         // per-move DOM query otherwise (mousemove is a hot path)
-        if (!dropdownElement || !isOpenRef.current || !hasItemsRef.current) return;
+        if (!dropdownElement || !isOpenRef.current || !hasItems) return;
         const pointer = { x: event.clientX, y: event.clientY };
         pointerRef.current = pointer;
         lastMouseEventRef.current = event.nativeEvent;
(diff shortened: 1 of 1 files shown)


packages/dropdown/src/Dropdown.tsx
```diff
@@ -342,40 +342,33 @@ function RootDropdown({
...
+    // handlers reading them are rebuilt whenever those props change — the
+    // React Compiler tracks them as dependencies of the memoized handler — so
+    // mirroring them bought nothing but a second source of truth.
     const isOpenRef = useRef(isOpen);
     const isOpeningRef = useRef(isOpening);
-    const keepOpenOnSubmitRef = useRef(keepOpenOnSubmit);
     const onCloseRef = useRef(onClose);
     const onOpenRef = useRef(onOpen);
-    const onSubmitItemRef = useRef(onSubmitItem);
-    const valueRef = useRef(valueIdentity);
```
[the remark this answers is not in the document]
acusti: Not stale — and the memoization is what prevents it, rather than what causes it. Checked two ways.

**The compiled output.** `handleRef` carries `allowCreate` in its memo guard:

```js
if ($[112] !== allowCreate || $[113] !== closeDropdown || $[114] !== disabled || $[115] !== handleActiveItem) {
    t44 = (ref) => { … };
}
…
const handleRef = t44;
```

So a changed `allowCreate` gives `handleRef` a new identity, React runs the old ref's cleanup (`inputElement.removeEventListener('input', handleInput)`) and attaches the new one with the fresh closure. The listener isn't attached once — it's re-attached exactly when a value it reads changes.

**Behaviorally**, flipping the prop after mount, with both controls:

| scenario | active item after typing `zz` |
|---|--:|
| `allowCreate={false}` throughout | **Alpha** |
| `allowCreate` from mount | **(none)** |
| flipped `false → true` at runtime | **(none)** |

The flipped case matches the always-true case, so the new value reached `isExactMatch`. Without the controls the third row would have been ambiguous — `(none)` could just mean nothing matched `zz`.

The general point is the reason this PR is safe at all: the compiler derives the dependency list from the closure, so a prop read inside `handleRef` is in its guard **by construction**, and if it couldn't preserve that memoization it would bail out rather than silently drop a dependency. The build reports zero bailouts, which is what makes reading props directly equivalent to the refs it replaces.

Worth noting this is the mirror image of a real bug we fixed in #414, where `handleBodyRef` **didn't** re-attach — precisely because nothing in its guard changed. Same mechanism, opposite conclusion; the guard contents are what decide it.

---

packages/dropdown/src/Dropdown.tsx
```diff
@@ -342,40 +342,33 @@ function RootDropdown({
...
+    //
+    // onOpen/onClose are called from an effect keyed on [isOpen] alone, which
+    // is what makes it fire on open/close transitions rather than whenever a
+    // consumer passes a fresh inline callback. Depending on them directly
+    // would mean listing them as dependencies and firing on identity changes.
+    //
+    // The rest of the props are read straight from the render closure. The
+    // handlers reading them are rebuilt whenever those props change — the
+    // React Compiler tracks them as dependencies of the memoized handler — so
+    // mirroring them bought nothing but a second source of truth.
```
```suggestion
    // The four values that still need mirroring, for two distinct reasons:
    // 1. isOpen/isOpening are read by the document-level listeners attached in
    //    handleRef
    // 2. onOpen/onClose are called from an effect keyed on [isOpen] alone, which
    //    is what makes it fire only on open/close transitions
```

acusti: In as written (2829368) — 16 lines to 5.

The numbered form is a real improvement over what I had: "for two distinct reasons" followed by two unlabelled paragraphs made you hold the count in your head, and the third paragraph then made it look like three.

What went: why reading state directly would churn the listeners, why depending on the callbacks directly would fire on identity changes, and the closing paragraph about the other props being tracked by the compiler. That last one was the least defensible — it explained the *absence* of code, on a line about the four refs that remain.

327 tests pass, zero compiler bailouts, lint/tsc/format clean.

---

PR #425, opened 2026-08-17
Give the selected item a forced-colors outline

acusti:
## Summary

- The persistent tint on the value-matching item (`--uktdd-body-bg-color-selected`) is a 6% `color-mix(in oklab, currentColor 6%, transparent)`, subtle by design.
- Under `forced-colors` mode (Windows High Contrast) that `color-mix()` still computes, but with next to nothing left to distinguish it from the surface — the one visual cue for the current selection disappears entirely for forced-colors users.
- The selected item now also gets an inset outline in `Highlight`, the same system color the hover/active state already uses, scoped to `@media (forced-colors: active)` so normal rendering is unchanged.

## How this was verified

happy-dom can't render media queries or compute real styles, and this repo has no visual-regression tooling, so this was checked against a real browser rather than assumed from spec reading:

- Extracted the actual compiled CSS from `dist/Dropdown.js` and rendered it in headless Chromium (Playwright + `page.emulateMedia({ forcedColors })`).
- Confirmed the *bug*: a literal author color (`background-color: red`) gets neutralized to `Canvas` under forced-colors, but `color-mix()` isn't neutralized the same way — it keeps evaluating with `currentColor` resolved to the forced foreground, so a 6% mix composites pixel-identical to the surrounding background.
- Confirmed the *fix*: screenshotted the selected item before/after in both modes — no visible change in normal mode (`outline: none`), a clear outline in forced-colors mode.
- Confirmed the outline still shows when an item is both selected and the current hover/active target (the two rules don't fight each other).
- Confirmed the nested `@media` survives lightningcss's CSS-nesting transform intact in the built bundle.

No automated test accompanies this — matching the existing convention in this file, where other pure-CSS decisions (dark-mode colors, tint derivation) ship with an explanatory comment rather than a vitest case, since the toolchain here can't exercise real CSS rendering.

## Test plan

- [x] `bun run build` — media query present in compiled output
- [x] `bun run test` (dropdown) — 163 passed
- [x] `bun run tsc` — clean
- [x] `bun run lint` — clean
- [x] `bun run format:check` — clean
- [x] Manual verification in headless Chromium under `forced-colors` emulation (see above)

https://claude.ai/code/session_01TPoB7JevJjKsuiEZjaw3UL

---

<a href="https://cubic.dev/pr/acusti/uikit/pull/425?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/dropdown-forced-colors-selected-outline.md
+++ b/.changeset/dropdown-forced-colors-selected-outline.md
@@ -4,14 +4,9 @@
 
 Give the selected item a forced-colors outline
 
-The persistent tint on `props.value`'s matching item is a 6% `color-mix()`,
+The persistent tint on `props.value`’s matching item is a 6% `color-mix()`,
 subtle by design. Under `forced-colors` mode (Windows High Contrast) that
 mix still computes, but leaves next to nothing to distinguish it from the
 surface, so the one visual cue for which item is selected disappears. The
-item now also gets an inset outline, scoped to
+item now also gets an inset outline in `currentColor`, scoped to
 `@media (forced-colors: active)` so normal rendering is unchanged.
-`currentColor` rather than a fixed system color, since this item can also
-be the active/hover target, whose own rule paints the same background
-`Highlight` — an outline fixed to that color would vanish into its own
-fill. `currentColor` instead tracks whichever color that rule leaves in
-effect for this element.
--- a/.changeset/dropdown-forced-colors-selected-outline.md
+++ b/.changeset/dropdown-forced-colors-selected-outline.md
@@ -8,6 +8,10 @@ The persistent tint on `props.value`'s matching item is a 6% `color-mix()`,
 subtle by design. Under `forced-colors` mode (Windows High Contrast) that
 mix still computes, but leaves next to nothing to distinguish it from the
 surface, so the one visual cue for which item is selected disappears. The
-item now also gets an inset outline in `Highlight`, the same system color
-the hover state already uses, scoped to `@media (forced-colors: active)` so
-normal rendering is unchanged.
+item now also gets an inset outline, scoped to
+`@media (forced-colors: active)` so normal rendering is unchanged.
+`currentColor` rather than a fixed system color, since this item can also
+be the active/hover target, whose own rule paints the same background
+`Highlight` — an outline fixed to that color would vanish into its own
+fill. `currentColor` instead tracks whichever color that rule leaves in
+effect for this element.
--- a/.changeset/dropdown-forced-colors-selected-outline.md
+++ b/.changeset/dropdown-forced-colors-selected-outline.md
@@ -1,0 +1,13 @@
+---
+'@acusti/dropdown': patch
+---
+
+Give the selected item a forced-colors outline
+
+The persistent tint on `props.value`'s matching item is a 6% `color-mix()`,
+subtle by design. Under `forced-colors` mode (Windows High Contrast) that
+mix still computes, but leaves next to nothing to distinguish it from the
+surface, so the one visual cue for which item is selected disappears. The
+item now also gets an inset outline in `Highlight`, the same system color
+the hover state already uses, scoped to `@media (forced-colors: active)` so
+normal rendering is unchanged.
--- a/packages/dropdown/README.md
+++ b/packages/dropdown/README.md
@@ -1185,14 +1185,8 @@ which case the component leaves selection ARIA entirely to you, as with any
 ARIA you set. The persistent tint on the selected item is themeable via the
 `--uktdd-body-bg-color-selected` custom property. The tint is subtle by
 design (a 6% mix), which `forced-colors` mode (Windows High Contrast)
-reduces further — a `color-mix()` still computes there, but with next to
-nothing left to distinguish it from the surface. The selected item also
-gets an outline in `currentColor`, scoped to
-`@media (forced-colors: active)` so it adds nothing outside that mode —
-`currentColor` rather than a fixed system color because this item can also
-be the hover/active target, whose own rule paints the same background
-`Highlight`; a fixed-color outline would vanish into its own fill, while
-`currentColor` tracks whichever color that rule leaves in effect.
+reduces further, so we give the selected item an outline in `currentColor`,
+scoped to `@media (forced-colors: active)`.
 
 ### The searchable trigger is a combobox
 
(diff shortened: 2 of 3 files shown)


packages/dropdown/src/Dropdown.css
```diff
@@ -196,6 +196,15 @@
 
     [aria-selected="true"] {
         background-color: var(--uktdd-body-bg-color-selected);
+        /* forced-colors mode drops the tint to a few percent of the forced
+           foreground, which is imperceptible against the forced background —
+           color-mix keeps computing, but with no reliable value left to mix
+           in. An inset outline in the same system color the hover state
+           already uses is a second, non-color-alone signal that survives */
+        @media (forced-colors: active) {
+            outline: 2px solid Highlight;
```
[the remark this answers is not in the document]
acusti: Valid, and my own verification had a gap that let it through. Fixed in 481bc8b, taking the `currentColor` option.

For the selected-only case I'd screenshotted the result. For selected+active I only checked that the two `getComputedStyle` calls returned non-error values and that the outline property was still `solid … 2px` — I didn't diff the actual color against the background, and didn't screenshot that combination. Doing both now confirms exactly what you found:

```
before: outline rgba(5, 0, 73, 0.8) solid 2px | background rgba(5, 0, 73, 0.8)   — identical
after:  outline rgb(255, 255, 255) solid 2px  | background rgba(5, 0, 73, 0.8)  — HighlightText vs Highlight
```

`currentColor` over a fixed `HighlightText`, since it self-adjusts to whichever rule is in effect on that element — `CanvasText` in the selected-only case (against the near-transparent tinted background), `HighlightText` once `[data-ukt-active]` also applies (against its `Highlight` background) — rather than hard-coding the second case and leaving the first on a color that happens to still work.

Re-verified both states with real screenshots this time: selected-only still shows a clear ring; selected+active now shows a white ring against the navy fill where it was previously one solid, undifferentiated block.

163 tests pass; tsc, lint, format clean.

---

.changeset/dropdown-forced-colors-selected-outline.md
```diff
@@ -0,0 +1,17 @@
...
+subtle by design. Under `forced-colors` mode (Windows High Contrast) that
+mix still computes, but leaves next to nothing to distinguish it from the
+surface, so the one visual cue for which item is selected disappears. The
+item now also gets an inset outline, scoped to
+`@media (forced-colors: active)` so normal rendering is unchanged.
+`currentColor` rather than a fixed system color, since this item can also
+be the active/hover target, whose own rule paints the same background
+`Highlight` — an outline fixed to that color would vanish into its own
+fill. `currentColor` instead tracks whichever color that rule leaves in
+effect for this element.
```
acusti: this last part is useful as context on the PR, but not for the changelog
```suggestion
```

.changeset/dropdown-forced-colors-selected-outline.md
```diff
@@ -0,0 +1,17 @@
+---
+'@acusti/dropdown': patch
+---
+
+Give the selected item a forced-colors outline
+
+The persistent tint on `props.value`'s matching item is a 6% `color-mix()`,
```
acusti:
```suggestion
The persistent tint on `props.value`’s matching item is a 6% `color-mix()`,
```

packages/dropdown/src/Dropdown.css
```diff
@@ -196,6 +196,20 @@
...
+        /* forced-colors mode drops the tint to a few percent of the forced
+           foreground, which is imperceptible against the forced background —
+           color-mix keeps computing, but with no reliable value left to mix
+           in. An inset outline is a second, non-color-alone signal that
+           survives. currentColor rather than a fixed Highlight: this item
+           can also be the active/hover target, whose rule below paints the
+           same background Highlight — an outline in that same color would
+           vanish into its own fill. currentColor instead tracks whichever
+           color that rule leaves in effect: CanvasText normally, or
+           HighlightText once the active rule’s `color` also applies */
```
acusti:
```suggestion
        /* forced-colors mode makes the selected item tint imperceptible, so
           we use an inset outline as a second, non-color-alone signal */
```

packages/dropdown/README.md
```diff
@@ -1183,7 +1183,16 @@ additionally receives `aria-selected="true"` (`aria-selected` isn’t valid
...
+`--uktdd-body-bg-color-selected` custom property. The tint is subtle by
+design (a 6% mix), which `forced-colors` mode (Windows High Contrast)
+reduces further — a `color-mix()` still computes there, but with next to
+nothing left to distinguish it from the surface. The selected item also
+gets an outline in `currentColor`, scoped to
+`@media (forced-colors: active)` so it adds nothing outside that mode —
+`currentColor` rather than a fixed system color because this item can also
+be the hover/active target, whose own rule paints the same background
+`Highlight`; a fixed-color outline would vanish into its own fill, while
+`currentColor` tracks whichever color that rule leaves in effect.
```
```suggestion
reduces further, so we give the selected item an outline in `currentColor`,
scoped to `@media (forced-colors: active)`.
```

acusti (commented):
looks good, but the comments and documentation needed simplifying

.changeset/dropdown-forced-colors-selected-outline.md
acusti: Dropped, in 2dd90dd. Fair — the `currentColor`-vs-`HighlightText` reasoning is exactly the kind of thing that belongs in the PR discussion (it's there, on the earlier thread) rather than repeated in what ships to consumers.

---

.changeset/dropdown-forced-colors-selected-outline.md
acusti: Fixed, in 2dd90dd — straight apostrophe to curly, matching the rest of the file's prose.

---

packages/dropdown/src/Dropdown.css
acusti: In as written, 2dd90dd — 11 lines to 2. The `currentColor` rationale that used to live here stays only in the PR thread now, not in the shipped comment.

---

packages/dropdown/README.md
acusti: In as written, 2dd90dd. Same trim as the CSS comment and changeset — states what the component does, leaves the "why `currentColor` and not a fixed color" reasoning to the PR thread rather than the docs.

---

acusti (commented):
LGTM

PR #426, opened 2026-08-20
Bump oxc-transform-react to 0.145.0, pin closure-capture fix

acusti:
## Summary

- Add an `it.fails` test pinning a known React Compiler bailout in `oxc-transform-react@0.144.0`: reassigning a destructured prop that a nested closure also captures triggers "Todo: Support destructuring of context variables" and "Immutability: This value cannot be modified" diagnostics — a gap that oxlint's `react/todo` and `react/immutability` rules currently miss (per [the linked upstream issue](https://github.com/oxc-project/oxc/issues)), even though `eslint-plugin-react-hooks`'s equivalent rules catch it.
- Bump `oxc-transform-react` to `0.145.0`, which adds support for this pattern — the tracking test flips from `it.fails` to a plain, enabled `it`.
- Drop the `sources: ['']` workaround in `index.ts` that neutralized `oxc-transform-react`'s own default `node_modules` substring filter (previously added so the plugin's own exact-path-segment filter stayed the sole authority on inclusion). 0.145.0's default filter behavior is adopted as-is rather than worked around, since the divergence — a directory name that merely contains "node_modules" as a substring — is a vanishingly rare edge case not worth diverging from upstream for. The `index.test.ts` case that pinned the old exact-segment guarantee is removed, since it was really pinning `oxc-transform-react`'s behavior rather than this plugin's own filter logic.

## Test plan

- [x] `bun run build`
- [x] `bun run test`
- [x] `bun run lint`
- [x] `bun run tsc`
- [x] `bun run format:check`

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vbcr4iu8ZgZ7PJZRet8Bok)_

<a href="https://cubic.dev/pr/acusti/uikit/pull/426?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/vite-plugin-react-compiler/src/upstream.test.ts
+++ b/packages/vite-plugin-react-compiler/src/upstream.test.ts
@@ -96,12 +96,13 @@ export function Counter({ step }: { step: number }) {
     });
 
     // reassigning a destructured prop identifier that a nested closure also
-    // captures bails the whole component out via two diagnostics —
+    // captures used to bail the whole component out via two diagnostics —
     // "Todo: Support destructuring of context variables" and "Immutability:
     // This value cannot be modified" — that eslint-plugin-react-hooks'
     // react-hooks/todo and react-hooks/immutability report but oxlint's
-    // react/todo and react/immutability miss
-    it.fails('compiles reassignment of a destructured prop captured by a closure', async () => {
+    // react/todo and react/immutability miss; fixed upstream in 0.145.0,
+    // pinned since
+    it('compiles reassignment of a destructured prop captured by a closure', async () => {
         const result = await transformCode(
             `
 export default function Repro({ value }: { value: string }) {
--- a/packages/vite-plugin-react-compiler/src/upstream.test.ts
+++ b/packages/vite-plugin-react-compiler/src/upstream.test.ts
@@ -94,4 +94,23 @@ export function Counter({ step }: { step: number }) {
         );
         expect(result?.code).not.toContain('react/compiler-runtime');
     });
+
+    // reassigning a destructured prop identifier that a nested closure also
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-react-compiler/src/upstream.test.ts
```diff
@@ -94,4 +94,24 @@ export function Counter({ step }: { step: number }) {
...
         expect(result?.code).not.toContain('react/compiler-runtime');
     });
+
+    // reassigning a destructured prop identifier that a nested closure also
+    // captures used to bail the whole component out via two diagnostics —
+    // "Todo: Support destructuring of context variables" and "Immutability:
+    // This value cannot be modified" — that eslint-plugin-react-hooks'
+    // react-hooks/todo and react-hooks/immutability report but oxlint's
+    // react/todo and react/immutability miss; fixed upstream in 0.145.0,
+    // pinned since
```
acusti:
```suggestion
    // This value cannot be modified"; fixed upstream in 0.145.0
```

acusti: Applied in 51e3d25.

---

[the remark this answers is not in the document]
acusti: Fixed in 51e3d25 — the dangling "pinned since" fragment is dropped entirely per the accepted suggestion above.

---

PR #427, opened 2026-08-20
Upgrade to native React Compiler build/lint and category-based oxlint config

acusti:
## Summary

- Upgrades oxlint to v1.79.0 and adopts its native React Compiler lint rules (22 granular `react/*` rules) in place of the old single `react/react-compiler` meta-rule.
- Replaces the Babel-based React Compiler transform (`@rolldown/plugin-babel` + `@vitejs/plugin-react`'s `reactCompilerPreset`) with `@vitejs/plugin-react@6.1.0`'s new native `compiler` option (`react({ compiler: {...} })`), backed by the same `oxc-transform-react` Rust bindings this repo's own `@acusti/vite-plugin-react-compiler` package wraps. Since `@vitejs/plugin-react` is already a plain, externally-published dependency every React package here uses, this drops the Babel pass entirely without adding a build-time dependency on `@acusti/vite-plugin-react-compiler` itself — that package is untouched and still published from this repo, just no longer used by this repo's own build. Because the native compiler runs across the Rust `oxc-transform-react` boundary, it can't accept callback-valued options like the Babel plugin's `logger`; compiler bailouts are instead caught statically at lint time by oxlint's new `react-compiler` rules.
- Switches `.oxlintrc.json` from an explicit per-rule allowlist to category-based config (`categories.correctness` / `categories.restriction` set to `"error"`), so new rules oxlint adds to those categories apply automatically instead of needing to be opted into one at a time. Rules that fall outside those two categories but are wanted (`no-shadow`, `capitalized-calls`, `hooks`, `memo-dependencies`, `rules-of-hooks`, `no-deriving-state-in-effects`) stay explicit, as do rules needing custom options or non-default severity.

## Category adoption details

The `restriction` off-list carries over the noise found adopting the same category split in a sibling codebase (opinionated style rules like `jsx-no-literals`, `explicit-function-return-type`, `no-param-reassign`, `no-non-null-assertion`, `require-test-timeout`), plus three entries specific to this codebase: `complexity`, `no-react-children`, and `no-clone-element` — Dropdown's compound-component internals (children traversal for label derivation, trigger cloning to inject ARIA props) and a couple of genuinely complex functions (a JSON parser, a dropdown state machine) trip these without being bugs.

Five `correctness`-category `jsx-a11y` rules also go on the off-list: `role-supports-aria-props`, `role-has-required-aria-props`, `interactive-supports-focus`, `no-noninteractive-tabindex`, and `prefer-tag-over-role`. Each flags a pattern that's core, tested behavior for Dropdown/Menubar rather than a mistake — consumer-authored `aria-selected`/`aria-disabled` on bare `li`/`button` items, pre-enhancement test fixtures for props Dropdown injects at render time, a menubar container that's correctly not itself a tab stop because it manages focus via roving tabindex on its children, and a test verifying Dropdown accepts an arbitrary non-button custom trigger element.

Everything else the categories surfaced gets fixed rather than suppressed: explicit type params on `vi.fn()` mocks, missing button `type` attributes, a couple of real type-safety gaps (an unsafe array `delete`, an unsound `String()` coercion, stringification of a too-wide `unknown` type), void-vs-undefined cleanup in two ref callbacks, and assorted small test/type fixes.

## Test plan

- [x] `bun run build` — all packages build clean, verified from a full `packages/*/dist` wipe
- [x] `bun run --filter '@acusti/' build` — verified standalone builds work for any package with no build-order dependency
- [x] `bun run lint` — clean
- [x] `bun run tsc` — clean across all workspaces
- [x] `bun run test` — 328/328 passing
- [x] `bun run format:check` — clean
- [x] Compiler output verified byte-identical to the previous setup (same memoization marker counts in `dropdown`/`css-value-input`, none in the opted-out `use-is-out-of-bounds`)

<a href="https://cubic.dev/pr/acusti/uikit/pull/427?utm_source=github" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"&gt;``</a>


vite.config.base.js
```diff
@@ -41,12 +40,13 @@ export const defineConfig = (options = {}) => {
...
+            // (see compilerOptions below) — kept out of this shared base
+            // config so packages that don’t opt in, including
+            // @acusti/vite-plugin-react-compiler itself, never need it
+            // pre-built
             ...(isReact ? [react()] : []),
-            // react: 'no-compiler' opts out of the React Compiler transform
-            // (for code that by design accesses refs during render)
-            ...(isReact === true
-                ? [babel({ presets: [reactCompilerPreset(compilerOptions)] })]
-                : []),
```
acusti, line 49: ok i see, so this looks like a bit of a regression. `css-value-input`  *was* being compiled already, but it didn’t need to duplicate that config inline, it could just pass `react: true` to enable that via `defineConfig` imported from this file. would be nice if we could preserve that behavior with the new vite-plugin-react-compiler

acusti (commented):
some initial thoughts/observations

packages/input-text/src/InputText.tsx
```diff
@@ -395,7 +396,7 @@ function hasNonPrimaryModifier(event: KeyboardEvent<InputElement>) {
 
 function isHeightEmpty(
     value?: null | number | string,
-): value is '' | 0 | null | typeof Infinity | undefined {
+): value is '' | null | typeof Infinity | undefined {
```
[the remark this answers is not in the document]
acusti: Checked this rather than applying the suggestion — it's a false positive. `typeof Infinity` resolves to plain `number` in TS (`Infinity` is declared `var Infinity: number`, not a literal type), which is exactly what the original oxlint error said when I made this change ("0 is overridden by number in this union type"). Confirmed in isolation with `tsc --strict`: after `isHeightEmpty(value)` narrows to true, `value === 0` still type-checks fine against the narrowed type, and `const x: typeof Infinity = 0` compiles — so `0` was already fully covered by `typeof Infinity`, removing it changed nothing about what the predicate allows.

Applying the suggested revert would just reintroduce the exact `typescript/no-redundant-type-constituents` violation this line was fixing (part of adopting oxlint's `correctness` category in this PR), so leaving it as-is.

---

vite.config.base.js
acusti: Agreed, and fixed — pushed in 33b644b. Moved the `@acusti/vite-plugin-react-compiler` wiring back into `defineConfig` in `vite.config.base.js`, dynamically imported so it's only resolved when `react` is exactly `true` (matching the original Babel setup's design). That keeps `@acusti/vite-plugin-react-compiler` itself, and any package that opts out, from ever needing it pre-built.

`use-is-out-of-bounds` now opts out via the `react: 'no-compiler'` sentinel the original Babel config supported, rather than by omitting a `plugins` entry. Reverted all 7 package `vite.config.js` files back to plain `react: true`. Verified the compiler still runs where expected: checked for `__COMPILER_RUNTIME`/memo-guard markers in the built output — present in `dropdown` and `css-value-input`, absent in `use-is-out-of-bounds`.

---

.changeset/css-value-input-first-react-compiler-release.md
```diff
@@ -0,0 +1,14 @@
...
+Compile with React Compiler
+
+CSSValueInput picked up `react: true` (enabling React Compiler) in its vite
+config after its last release, so this is the first published build
+compiled with React Compiler — via @acusti/vite-plugin-react-compiler,
+which runs the compiler through oxc-transform-react (the native Node
+bindings for the oxc project's Rust port), rather than the Babel-based
+pipeline the sibling packages in this repo are moving off of. Output is
+verified equivalent to the Babel-based compiler: React Compiler memoization
+is present with zero compiler bailouts. No API or behavior changes.
```
acusti: this if for external consumers of the library, skip any inside baseball stuff. it should be written with just the info that is useful to someone using or considering using this package

packages/dropdown/src/Dropdown.tsx
```diff
@@ -443,7 +443,10 @@ function RootDropdown({
     };
 
     // Clear the disclosure intent timer on unmount so it can’t fire against
-    // unmounted DOM or dispatch submenu callbacks after teardown
+    // unmounted DOM or dispatch submenu callbacks after teardown. Deliberately
+    // unmount-only: clearDisclosureTimer only touches refs, so leaving it out
+    // of deps is safe and re-running it on every render is not desired.
+    // oxlint-disable-next-line react/exhaustive-effect-dependencies
```
acusti, line 449: if clearDisclosureTimer only gets called from within an effect, we should define it via useEffectEvent, and then it doesn't even need refs and gets a closure that gives it access to the latest of all props and state and such. but if that doesn't work, i'm ok with the disable-next-line

packages/dropdown/src/Dropdown.tsx
```diff
@@ -478,7 +481,8 @@ function RootDropdown({
     };
 
     // Clear the safe-area give-up timer on unmount so it can’t fire against
-    // unmounted DOM
+    // unmounted DOM. Deliberately unmount-only, see clearDisclosureTimer above.
+    // oxlint-disable-next-line react/exhaustive-effect-dependencies
```
acusti, line 485: same here as above

packages/dropdown/src/Dropdown.tsx
```diff
@@ -590,7 +594,8 @@ function RootDropdown({
     };
 
     // Clear props.openOnHover’s close timer on unmount so it can’t fire against
-    // unmounted DOM
+    // unmounted DOM. Deliberately unmount-only, see clearDisclosureTimer above.
+    // oxlint-disable-next-line react/exhaustive-effect-dependencies
```
acusti, line 598: same here as above twice

acusti (commented):
next set of comments

.changeset/css-value-input-first-react-compiler-release.md
acusti: Fair, rewrote it (and the sibling `adopt-native-react-compiler.md` changeset, since it had the same problem — mentioned `oxc-transform-react`, the Rust port, the Babel pipeline being replaced, which other packages share the change). Both now just say the build runs through React Compiler, output is verified equivalent, no API/behavior changes. Pushed in 3799e1c.

---

packages/dropdown/src/Dropdown.tsx
acusti: Checked — it doesn't only get called from an effect, so keeping the disable-next-line. `clearDisclosureTimer` is also called from `syncSubmenuDisclosure` (invoked from pointer tracking and `closeDropdown`), `handleSubmitItem`, and the ArrowRight/ArrowLeft keyboard handling — all regular event handlers, not effects. Since Effect Events can only be called from inside an Effect, wrapping it in `useEffectEvent` would just move the violation from "missing dep" to "calling an Effect Event outside an Effect."

---

packages/dropdown/src/Dropdown.tsx
acusti: Same story — `clearSafeAreaTimer` is also called from `trackSafeArea` (mousemove), `commitPointerTarget`, `handleMouseOut`, and `closeDropdown`, all regular event handlers rather than effects, so it's not a fit for `useEffectEvent` either.

---

packages/dropdown/src/Dropdown.tsx
acusti: Same as the other two — `clearHoverCloseTimer` is also called from `handleDropdownMouseEnter` and `closeDropdown`, regular event handlers, not effects.

---

.changeset/css-value-input-first-react-compiler-release.md
```diff
@@ -0,0 +1,9 @@
+---
+'@acusti/css-value-input': patch
+---
+
+Compile with React Compiler
+
+This package's build now runs through React Compiler, so its internals are
```
acusti:
```suggestion
This package’s build now runs through React Compiler, so its internals are
```

packages/css-value-input/package.json
```diff
@@ -44,6 +44,7 @@
         "clsx": "^2"
     },
     "devDependencies": {
+        "@acusti/vite-plugin-react-compiler": "workspace:^",
```
acusti: if we’re keeping react compiler usage within `vite.config.base.ts` at the root of the repo, should we still add the devDeps in each workspace?

package.json
```diff
@@ -46,19 +46,16 @@
         "@changesets/cli": "^3"
     },
     "devDependencies": {
-        "@babel/core": "^7.29.7",
-        "@rolldown/plugin-babel": "^0.2.3",
```
acusti, line 50: seems like if we’re replacing these at the root, we should also add a reference to `@acusti/vite-plugin-react-compiler`?

acusti (commented):
not sure if the existing babel react-compiler implementation was sound, but assuming it was, we should continue to follow it in our version of that implementation

.changeset/css-value-input-first-react-compiler-release.md
acusti: Applied, pushed in b2c1f0f.

---

packages/css-value-input/package.json
acusti: Good question — checked what the Babel setup did, since it's the closest precedent: `@babel/core`, `@rolldown/plugin-babel`, and `babel-plugin-react-compiler` only ever lived in the root `package.json`, never duplicated per-package, because only `vite.config.base.js` imported them. Same logic applies now that the wiring is back in `defineConfig` — removed `@acusti/vite-plugin-react-compiler` from this package (and the other 6 that pass `react: true`) and added it to root instead. Pushed in b2c1f0f.

`@acusti/uikit-docs` keeps its own entry, since its Storybook config imports the plugin directly rather than going through `defineConfig`.

---

package.json
acusti: Yep — added `@acusti/vite-plugin-react-compiler` to root, matching where the Babel deps it replaces used to live (they were root-only too). Same commit as the reply below, b2c1f0f.

---

packages/css-value-input/package.json
acusti: Correcting my last reply here — that broke CI. The Babel-only-at-root precedent doesn't actually transfer: Babel's deps were plain npm packages needing no build step, but `@acusti/vite-plugin-react-compiler` is a workspace package that does. Vite's config-file bundler resolves a literal `import()` target even when the branch that runs it is unreachable, so with the dep only at root, *every* package's build (not just React ones) needed it pre-built just to load `vite.config.base.js` — and nothing in the workspace build-order script (which topologically sorts by declared dependencies) knew to build it first anymore, since I'd removed it from here. Caught this because CI's `bun install --frozen-lockfile` starts from no pre-existing `dist/`, which my local rebuilds hadn't been doing.

Fix (pushed in 33ae145): made the import specifier computed rather than literal, so vite's bundler only resolves it for packages that actually reach that branch (`react: true`), and re-added the devDependency here — it's a real build-order requirement even though this package doesn't import the plugin itself. Verified with a full `packages/*/dist` wipe + rebuild matching CI's cold-checkout conditions.

---

package.json
acusti: This one's still right — root genuinely only needs it here, since `vite.config.base.js` (the only place that imports it at the root level) lives at root. See my correction on the css-value-input thread for what was wrong: the packages that pass `react: true` need it back as a devDependency too, for build-order reasons the Babel precedent didn't have to deal with.

---

packages/use-is-out-of-bounds/vite.config.js
```diff
@@ -1,3 +1,8 @@
 import { defineConfig } from '../../vite.config.base.js';
 
-export default defineConfig({ entry: ['src/useIsOutOfBounds.ts'], react: 'no-compiler' });
+// react: 'no-compiler' opts out of the React Compiler transform for code
+// that by design accesses refs during render
```
acusti:
```suggestion
// react: 'no-compiler' opts out of the React Compiler transform for this
// custom hook that by design accesses refs during render
```

acusti: Applied, pushed in 38fe5c6.

---
_Generated by [Claude Code](https://claude.ai/code)

---

packages/use-keyboard-events/src/handlers.ts
```diff
@@ -35,18 +35,18 @@ export function addHandler({
     );
 
     const handlersByPriority = handlersData[eventType];
-    handlersByPriority[priority] ??= new Set();
-
-    handlersByPriority[priority].add(handler);
+    const handlers = (handlersByPriority[priority] ??= new Set());
+    handlers.add(handler);
```
acusti: this changes makes this hard for me to read. maybe we should just do something like `if (handlersByPriority[priority] == null) { handlersByPriority[priority] = new Set(); } handlersByPriority[priority].add(handler);`

packages/use-keyboard-events/src/handlers.ts
```diff
@@ -35,18 +35,18 @@ export function addHandler({
...
-        if (handlersByPriority[priority] == null) return;
+        const handlersAtPriority = handlersByPriority[priority];
+        if (handlersAtPriority == null) return;
 
-        handlersByPriority[priority].delete(handler);
-        if (!handlersByPriority[priority].size) {
-            delete handlersByPriority[priority];
+        handlersAtPriority.delete(handler);
+        if (!handlersAtPriority.size) {
+            handlersByPriority[priority] = undefined;
```
acusti, line 49: this means that `handlersByPriority` will still have an entry at the `priority`. do we iterate all of `handlersByPriority` at any point where keeping an empty `priority` around would uselessly slow down iteration?

packages/use-keyboard-events/src/handlers.ts
acusti: Your version conflicts with `typescript/prefer-nullish-coalescing`, an existing rule in this repo's oxlint config (predates this PR) that wants `??=` over an if-null-then-assign — so applying it as written would trade one lint error for another.

Went with a middle ground instead: dropped the `const handlers = (… ??= …)` capture (probably the actual confusing part — an assignment used as an expression value) but kept `??=` as a bare statement, then re-indexed for `.add()`. Pushed in ba7fd3b.

---

packages/use-keyboard-events/src/handlers.ts
acusti: No — checked where `handlersByPriority` actually gets read. The only iteration site is `handleKeyboardEvent` (line 145 area), and it's a manual `while (index--)` loop over every index from `.length` down to `0`, not `forEach`/`map`/`filter`. Hole-skipping only applies to those callback-based array methods; a plain indexed loop visits every slot regardless, and this one already does a `!= null` check per slot either way — so a hole from `delete` and an explicit `undefined` are indistinguishable to it. The trailing-trim loop right below the assignment (shrinking `.length` when trailing slots are empty) also treats both the same way, so `.length` doesn't grow unbounded either.

---

vite.config.base.js
```diff
@@ -16,6 +15,25 @@ export const defineConfig = (options = {}) => {
...
+    const reactCompilerPluginSpecifier = '@acusti/vite-plugin-react-compiler';
+    const reactCompilerPlugins =
+        isReact === true
+            ? [
+                  (await import(reactCompilerPluginSpecifier)).default({
+                      reactCompiler: compilerOptions,
+                  }),
+              ]
+            : [];
+
```
acusti: this feels like a hacky exploit of a weakness in static analysis (creating `reactCompilerPliuginSpecifier` as a static string const that we then dynamically import to prevent vite from being able to identify what it is). maybe we should just add `@acusti/vite-plugin-react-compiler` as a regular dependency and point it at `0.2.0` instead of trying to use the workspace’s built on demand version

vite.config.base.js
```diff
@@ -57,27 +73,12 @@ export const defineConfig = (options = {}) => {
 };
 
 // React Compiler https://github.com/reactwg/react-compiler/discussions/36#discussioncomment-11285011
+// (compiler bailouts are caught by oxlint's react-compiler rules instead of
+// a runtime logger: @acusti/vite-plugin-react-compiler doesn't support
+// callback-valued options like `logger` since they can't cross the native
```
acusti:
```suggestion
// a runtime logger: @acusti/vite-plugin-react-compiler doesn’t support
// callback-valued options like `logger` since they can’t cross the native
```

acusti: Applied, pushed in e5c8c69.

---

vite.config.base.js
acusti: Tried it — doesn't work, for a reason worth knowing about: bun workspaces always claim priority for a package name that matches a workspace member, regardless of the declared version range. Changed root's entry to `^0.2.0` and even tried the `npm:@acusti/vite-plugin-react-compiler@^0.2.0` alias to force registry resolution; both times `node_modules/@acusti/vite-plugin-react-compiler` stayed a symlink to `packages/vite-plugin-react-compiler` (confirmed with a forced clean reinstall, not just a stale cache). There's no version range that makes bun prefer the registry over a same-named workspace member.

So I looked for a way to keep the import literal without that trick. Tried giving `vite-plugin-react-compiler`'s own `vite.config.js` a standalone config that doesn't import the shared `defineConfig` at all (avoiding self-reference), which would let the root import stay a plain `import reactCompiler from '@acusti/vite-plugin-react-compiler'`. That's not enough either — proved it by wiping every package's `dist/` and running a full cold build: `@acusti/aws-signature-v4`, which has nothing to do with React, failed with the identical resolution error. Every package's `vite.config.js` goes through the shared `vite.config.base.js`, so a literal top-level import there gets eagerly resolved by Vite's config-file bundler for *any* consumer of that file, not just React ones or `vite-plugin-react-compiler` itself. Isolating just the self-referential package doesn't fix that — it'd take standalone configs for all 9 non-React packages to dodge it, which is a lot more duplication than the one computed-specifier line.

So the computed specifier is staying — it's not dodging a static-analysis weakness for its own sake, it's the one thing that actually survives a genuinely cold build across the whole workspace (verified twice now with a full `packages/*/dist` wipe). Reverted the `vite.config.base.js`/`.d.ts` changes back to what's already on this branch (`33ae145`) — nothing new to push there.

---

acusti: Update: went with the build-order fix after all. Pushed in 129ebb0:

- `vite.config.base.js` is back to a plain literal static import, no computed specifier.
- `scripts/run-workspaces.mjs` forces `@acusti/vite-plugin-react-compiler` to build first, unconditionally — nothing in the topological sort could express "after itself" for a package that's indirectly self-referential through the shared config every other package also loads.
- `vite-plugin-react-compiler`'s own `vite.config.js` no longer goes through the shared `defineConfig` (it'd still need its own `dist/` pre-built regardless of build order otherwise), so it duplicates the small amount of build/test config it needs directly.
- The 7 `react: true` packages no longer declare `@acusti/vite-plugin-react-compiler` as a `devDependency` — that only existed to give the topological sort an edge to work with, which the hardcoded order now makes moot.

Verified the same way: full `packages/*/dist` wipe, `bun run build`. The plugin builds first, and `aws-signature-v4` (and everything else) resolves cleanly afterward.

---

vite.config.base.js
```diff
@@ -1,7 +1,7 @@
 import { fileURLToPath } from 'node:url';
 
-import babel from '@rolldown/plugin-babel';
-import react, { reactCompilerPreset } from '@vitejs/plugin-react';
+import reactCompiler from '@acusti/vite-plugin-react-compiler';
```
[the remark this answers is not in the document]
acusti: Confirmed this reproduces — `rm -rf packages/*/dist && bun run --filter '@acusti/aws-signature-v4' build` on a cold checkout fails with the same resolution error, since that bypasses `scripts/run-workspaces.mjs`'s hardcoded build-first order entirely.

This is a known, deliberate tradeoff, not a new bug: the previous version of this line used a dynamic import with a computed specifier specifically to keep non-React packages immune to this, but that got replaced with this plain literal import on purpose (see the thread above at line 36 — now moved, was `vite.config.base.js:36`) because a dynamic computed-specifier import that only exists to defeat a bundler's static analysis was judged not worth carrying long-term. The accepted cost is: `bun run build` (which forces the compiler package first) stays safe on a cold checkout, but a standalone `bun run --filter` build of a non-React package now needs that build, or at least the compiler package's `dist/`, to already exist — matching how a standalone `--filter` build of `dropdown` already needs `@acusti/use-keyboard-events` pre-built today, since it's a real dependency.

---

scripts/run-workspaces.mjs
```diff
@@ -49,7 +49,9 @@ for (let index = 0; index < rawArgs.length; index += 1) {
     process.exit(1);
 }
 
-const rootManifest = JSON.parse(await fs.readFile(path.join(rootDir, 'package.json'), 'utf8'));
+const rootManifest = JSON.parse(
+    await fs.readFile(path.join(rootDir, 'package.json'), 'utf8'),
+);
```
acusti, line 54: i can't tell, is there a difference here? if not, let's not change the newline formatting for no clear reason

scripts/run-workspaces.mjs
```diff
@@ -96,10 +98,17 @@ for (const workspaceDir of workspaceDirs) {
     }
 }
 
-const selectedWorkspaces = workspaces.filter((workspace) => !excludes.has(workspace.name));
+const selectedWorkspaces = workspaces.filter(
+    (workspace) => !excludes.has(workspace.name),
+);
```
acusti, line 103: same here

scripts/run-workspaces.mjs
```diff
@@ -96,10 +98,17 @@ for (const workspaceDir of workspaceDirs) {
...
 const selectedNames = new Set(selectedWorkspaces.map((workspace) => workspace.name));
 
-const dependencySections = ['dependencies', 'devDependencies', 'peerDependencies', 'optionalDependencies'];
+const dependencySections = [
+    'dependencies',
+    'devDependencies',
+    'peerDependencies',
+    'optionalDependencies',
+];
```
acusti, line 111: same here

scripts/run-workspaces.mjs
```diff
@@ -141,7 +150,26 @@ if (orderedNames.length !== selectedWorkspaces.length) {
     process.exit(1);
 }
 
-const workspaceByName = new Map(selectedWorkspaces.map((workspace) => [workspace.name, workspace]));
+// @acusti/vite-plugin-react-compiler is imported by every package's shared
+// vite.config.base.js (see that file for why), including its own — nothing
+// can express "after itself" in the dependency graph the sort above reads,
+// so its build order has to be forced here instead. Move it to the front
+// unconditionally so its dist/ always exists before any other package's
+// vite config gets loaded.
```
acusti:
```suggestion
// @acusti/vite-plugin-react-compiler is imported by every package’s shared
// vite.config.base.js (see that file for why), including its own — nothing
// can express “after itself” in the dependency graph the sort above reads,
// so its build order has to be forced here instead. Move it to the front
// unconditionally so its dist/ always exists before any other package’s
// vite config gets loaded.
```

acusti (commented):
this looks way better, thx. just some nits

scripts/run-workspaces.mjs
acusti: No semantic difference — just oxfmt's line-wrapping. Tried reverting it, but it doesn't stick: the pre-commit hook runs oxfmt on every staged file matching its extensions (not scoped to the same include list `bun run format`/`format:check` use, which don't cover `./scripts/`), so as soon as this file is staged in any commit it gets reformatted back to this. My first attempt at reverting it actually got silently undone by the hook during that same commit — I caught it by checking what actually landed vs. what I'd staged.

So this isn't something I chose, and I can't make it stick without bypassing the hook (which I won't do without you asking for it) or excluding this file from oxfmt entirely (a bigger change than the 3-line nit here). Leaving it as oxfmt formats it.

---

scripts/run-workspaces.mjs
acusti: Same as the reply above — this is the pre-commit hook's oxfmt pass, not a change I made, and it doesn't stick if reverted since any commit touching this file re-triggers it.

---

scripts/run-workspaces.mjs
acusti: Same as the reply above.

---

scripts/run-workspaces.mjs
acusti: Applied, pushed in 24636a1.

---

packages/input-text/src/InputText.tsx
```diff
@@ -395,7 +396,7 @@ function hasNonPrimaryModifier(event: KeyboardEvent<InputElement>) {
...
     return value == null || value === '' || value === 0 || value === Infinity;
 }
```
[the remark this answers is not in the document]
acusti: Already verified as a false positive earlier in this PR (see the resolved thread above on this same line) — `typeof Infinity` resolves to plain `number` in TS, not a literal excluding other numbers, so `0` was already covered and removing it changed nothing about what the predicate narrows to. Confirmed with an isolated `tsc --strict` check.

---

packages/use-keyboard-events/src/handlers.ts
```diff
@@ -36,17 +36,17 @@ export function addHandler({
 
     const handlersByPriority = handlersData[eventType];
     handlersByPriority[priority] ??= new Set();
-
     handlersByPriority[priority].add(handler);
     handlersData.config.set(handler, { ignoreUsedKeyboardEvents, priority });
```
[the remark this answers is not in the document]
acusti: Checked — `bun run tsc` is clean on this file as written. TS does narrow `handlersByPriority[priority]` after the `??=` here (no intervening statements that could invalidate it between the assignment and the `.add()` call), so there's no actual type error to fix. A local variable would be the safer pattern in general, but this specific case already type-checks without one.

---

scripts/run-workspaces.mjs
```diff
@@ -141,7 +150,26 @@ if (orderedNames.length !== selectedWorkspaces.length) {
...
+const buildFirstNames = ['@acusti/vite-plugin-react-compiler'];
+
+for (const name of buildFirstNames) {
+    const index = orderedNames.indexOf(name);
+
+    if (index > 0) {
+        orderedNames.splice(index, 1);
+        orderedNames.unshift(name);
+    }
+}
```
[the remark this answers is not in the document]
acusti: Good catch, valid — fixed. Pushed in a8d4a9f: `--exclude @acusti/vite-plugin-react-compiler` now errors out explicitly instead of silently breaking every other package's build.

---

acusti (commented):
LGTM

vite.config.base.js
```diff
@@ -57,27 +57,11 @@ export const defineConfig = (options = {}) => {
 };
 
 // React Compiler https://github.com/reactwg/react-compiler/discussions/36#discussioncomment-11285011
+// (compiler bailouts are caught by oxlint's react-compiler rules instead of
+// a runtime logger: callback-valued options like `logger` aren’t supported
+// since they can’t cross the native oxc-transform-react boundary)
```
acusti: let’s make sure this is covered in the PR description and then drop this (i try to avoid code comments that describe how the code used to be, rather than how the code is)

vite.config.base.js
```diff
@@ -41,11 +40,12 @@ export const defineConfig = (options = {}) => {
...
         ...(css ? { css } : {}),
         plugins: [
-            ...(isReact ? [react()] : []),
             // react: 'no-compiler' opts out of the React Compiler transform
-            // (for code that by design accesses refs during render)
-            ...(isReact === true
-                ? [babel({ presets: [reactCompilerPreset(compilerOptions)] })]
+            // (for code that by design accesses refs during render); the
+            // compiler runs via @vitejs/plugin-react's native compiler
+            // option, backed by oxc-transform-react
```
acusti: no need to call out using the @vitejs/plugin-react native support
```suggestion
            // (for code that by design accesses refs during render)
```

vite.config.base.js
acusti: Added the `logger`/native-boundary detail to the PR description (second bullet under Summary) and dropped the comment. Pushed in 155cd2c.

---

vite.config.base.js
acusti: Applied verbatim. Pushed in 155cd2c.

---

PR #435, opened 2026-08-28
Reconcile the menubar’s tab stop only when its holder actually leaves

### LINKED ISSUE #431: dropdown: Menubar reconciles on every member re-render, not just on unmount
`registerMember`’s cleanup bumps the reconcile reducer unconditionally, but members re-register on **every** render — so every state change in any menubar member costs the `Menubar` an extra state update and re-render that does nothing.

## Where

`packages/dropdown/src/Menubar.tsx`, `registerMember`:

```ts
return () => {
    membersRef.current.delete(member);
    // say the set has settled, so the effect above can pick
    // up a tab stop stranded on an unmounted member
    reconcile();
    // the tab stop is deliberately not released here: members
    // re-register on every render, and React runs every cleanup
    // before any setup, so releasing would hand it to whichever
    // member re-registers first rather than back to the holder
};
```

The second comment states the premise directly: the register effect is deps-less, so members re-register on every render. That means this cleanup runs on every member render, not only on unmount, and `reconcile()` fires each time.

## Effect

Any member state change — `isOpen`/`isOpening` toggling, hover open/close — triggers cleanup → `reconcile()` → a `Menubar` re-render whose effect immediately early-returns:

```ts
const owner = members.find((member) => member.element === tabbableElement);
if (owner && canHoldTabStop(owner)) return;
```

So it's one wasted `Menubar` render per member render, for the lifetime of the bar. Purely a performance/noise issue — the behavior is correct, and the reconcile is genuinely needed on real unmounts (a tab stop stranded on a departing member).

## Possible fix

Bump only when the departing member could actually affect the tab stop — e.g. when `member.element === tabbableElement` — so steady-state re-registration doesn't schedule anything. Worth confirming that still covers the stranded-tab-stop case the reconcile exists for.

Found during a deep review of `main` (`a48d903`).

acusti:
Fixes #431.

`registerMember`'s cleanup bumped the reconcile reducer unconditionally. Because the register effect in `Dropdown.tsx` is deps-less (`useEffect(() => {…})`), members re-register on **every** render, so that cleanup runs far more often than a member actually leaves — and each bump scheduled a `Menubar` state update whose effect then almost always early-returned, the tab stop still being held by an eligible member.

The change splits two situations the old code conflated:

- **A genuine departure.** A member that merely re-rendered re-registers during the same commit, so the decision is deferred until the commit settles and the bump fires only if that element never came back.
- **An eligibility transition** (the holder becomes `disabled`, or stops being a menu). Settled on the registration side, where the freshly registered member reflects the props that changed: it returns `null` and the effect rehomes the tab stop. Every other re-registration returns `current` untouched, so React bails out instead of re-rendering the bar.

## Measured, not assumed

Counting `Menubar` commits with a `<Profiler>`, opening and closing the tab stop holder's own menu three times — the common interaction, which re-registers the holder repeatedly while it stays eligible:

| | Menubar commits |
|---|---|
| before | 19 |
| after | 9 |

## Two corrections along the way

Both from measuring or reproducing rather than reasoning, and both worth reading before reviewing the diff:

1. **#431 overstated the cost.** It predicted "one wasted render per member render". React batches the bumps within a commit, so the real cost is roughly one extra render per commit in which members re-render.
2. **The first two attempts were each wrong, and cubic caught both.**
   - `member.element === tabbableElement` alone (b676aa7) left the common case at **19 commits — identical to no fix at all**. The 27 → 12 figure quoted in an earlier description came only from *other* members churning during hover-switching.
   - Adding `&& !member.element.isConnected` (a6ae04c) regressed `<Activity mode="hidden">`: React unmounts a hidden subtree's effects while leaving its DOM in place, so the holder deregistered with its trigger still connected, the bump was skipped, and the tab stop stranded on an unreachable trigger. Reproduced on React 19.2.8 before fixing. The unconditional bump had handled that correctly.

   The lesson in both: the question is "did this registration go away?", and neither "is it the holder?" nor "is its element still in the document?" answers it. 8ce1ca3 stops inspecting the DOM and lets the commit settle instead, which covers unmount, hidden subtrees, and any other teardown without a special case for each.

## Correctness

Two tests for cases the existing suite didn't reach, each failing on the attempt it rules out:

- **the holder becoming `disabled` without unmounting** — fails on `isConnected`-style narrowing, which is what forced the eligibility half of the design.
- **the holder hidden with `<Activity>`** — fails on a6ae04c, passes on 8ce1ca3.

## A gap worth naming

There is no CI guard against a future regression to unconditional bumping. I wrote a `<Profiler>` render-count test and removed it, because it can't be reliable here:

| config | before | after |
|---|---|---|
| `packages/dropdown/vite.config.js` (`react: true`, compiler on) | 19 | 9 |
| root `vitest.config.js` (no react plugin, compiler off) | 19 | **18** |

CI runs `bun run test` from the root, where the compiler is off and the gap is one commit — too tight for a threshold that wouldn't flake on a React bump. Without the compiler the bar re-renders for other reasons that mask the reconcile bumps; with it (as in the shipped build) they're the dominant remaining cost. Covering this properly would need a compiler-enabled test project rather than a threshold in the shared suite; happy to open a follow-up issue if that's wanted.

Full repo `build`, `test` (392 passed / 24 files), `lint`, `tsc`, and `format:check` pass, as does the compiled path (`bun run --filter '@acusti/dropdown' test`, 167 passed). Includes a patch changeset.

Left as three commits so cubic's reviews stay anchored to what they reviewed — squash on merge.

https://claude.ai/code/session_01An3mqqrhHB7b7FCLt6muyK

### DIFF
--- a/.changeset/dropdown-menubar-reconcile-only-for-holder.md
+++ b/.changeset/dropdown-menubar-reconcile-only-for-holder.md
@@ -2,16 +2,19 @@
 '@acusti/dropdown': patch
 ---
 
-Reconcile the menubar’s tab stop only when its holder leaves
+Reconcile the menubar’s tab stop only when its holder actually leaves
 
 Menubar members re-register on every render, so the cleanup that removes a
 member from the set runs far more often than a member actually leaves. It
 bumped the reconcile reducer unconditionally, which scheduled a Menubar
-state update and re-render for every member render — and the effect it woke
-almost always early-returned, because the tab stop was still held by an
-eligible member.
+state update and re-render each time, and the effect it woke almost always
+early-returned because the tab stop was still held by an eligible member.
 
-Only the holder going (or becoming ineligible, which re-registers it here
-too) can strand the tab stop, so only that case now reconciles. On a
-four-member bar, engaging it and hover-switching around it three times went
-from 27 Menubar renders to 12.
+Only the holder really going can strand the tab stop, and by the time the
+cleanup runs React has already detached its element — which is what
+distinguishes a departure from a re-render. Losing eligibility without
+leaving (a holder that becomes disabled, or stops being a menu) is now
+settled where the member re-registers, against the props that changed.
+
+Measured on the compiled build, opening and closing the tab stop holder’s
+own menu three times: 19 Menubar commits before, 9 after.
--- a/.changeset/dropdown-menubar-reconcile-only-for-holder.md
+++ b/.changeset/dropdown-menubar-reconcile-only-for-holder.md
@@ -10,11 +10,12 @@ bumped the reconcile reducer unconditionally, which scheduled a Menubar
 state update and re-render each time, and the effect it woke almost always
 early-returned because the tab stop was still held by an eligible member.
 
-Only the holder really going can strand the tab stop, and by the time the
-cleanup runs React has already detached its element — which is what
-distinguishes a departure from a re-render. Losing eligibility without
-leaving (a holder that becomes disabled, or stops being a menu) is now
-settled where the member re-registers, against the props that changed.
+Only the holder really going can strand the tab stop, and a member that
+merely re-rendered re-registers during the same commit — so the bump now
+waits for the commit to settle and fires only if that member never came
+back. Losing eligibility without leaving (a holder that becomes disabled,
+or stops being a menu) is settled where the member re-registers, against
+the props that changed.
 
 Measured on the compiled build, opening and closing the tab stop holder’s
 own menu three times: 19 Menubar commits before, 9 after.
--- a/.changeset/dropdown-menubar-reconcile-only-for-holder.md
+++ b/.changeset/dropdown-menubar-reconcile-only-for-holder.md
@@ -1,0 +1,17 @@
+---
+'@acusti/dropdown': patch
+---
+
+Reconcile the menubar’s tab stop only when its holder leaves
+
+Menubar members re-register on every render, so the cleanup that removes a
+member from the set runs far more often than a member actually leaves. It
+bumped the reconcile reducer unconditionally, which scheduled a Menubar
+state update and re-render for every member render — and the effect it woke
+almost always early-returned, because the tab stop was still held by an
+eligible member.
+
+Only the holder going (or becoming ineligible, which re-registers it here
+too) can strand the tab stop, so only that case now reconciles. On a
+four-member bar, engaging it and hover-switching around it three times went
+from 27 Menubar renders to 12.
--- a/.changeset/dropdown-menubar-reconcile-only-for-holder.md
+++ b/.changeset/dropdown-menubar-reconcile-only-for-holder.md
@@ -2,20 +2,4 @@
 '@acusti/dropdown': patch
 ---
 
-Reconcile the menubar’s tab stop only when its holder actually leaves
-
-Menubar members re-register on every render, so the cleanup that removes a
-member from the set runs far more often than a member actually leaves. It
-bumped the reconcile reducer unconditionally, which scheduled a Menubar
-state update and re-render each time, and the effect it woke almost always
-early-returned because the tab stop was still held by an eligible member.
-
-Only the holder really going can strand the tab stop, and a member that
-merely re-rendered re-registers during the same commit — so the bump now
-waits for the commit to settle and fires only if that member never came
-back. Losing eligibility without leaving (a holder that becomes disabled,
-or stops being a menu) is settled where the member re-registers, against
-the props that changed.
-
-Measured on the compiled build, opening and closing the tab stop holder’s
-own menu three times: 19 Menubar commits before, 9 after.
+Halve the renders a Menubar does as its menus open and close
--- a/packages/dropdown/src/Menubar.test.tsx
+++ b/packages/dropdown/src/Menubar.test.tsx
@@ -327,6 +327,54 @@ describe('@acusti/dropdown Menubar', () => {
                 screen.getByRole('menuitem', { name: 'Edit' }).getAttribute('tabindex'),
             ).toBe('0');
         });
+
+        it('hands the tab stop on when the holder becomes disabled', async () => {
+            const user = userEvent.setup();
+            // Like the unmount above, this is driven by the members’ own state,
+            // so Menubar never re-renders on its own. The holder stops being
+            // eligible without leaving the set, so it’s the holder’s own
+            // re-registration that has to schedule the reconciliation.
+            const Members = () => {
+                const [isFileDisabled, setIsFileDisabled] = useState(false);
+                return (
+                    <>
+                        <button onClick={() => setIsFileDisabled(true)} type="button">
+                            disable File
+                        </button>
+                        <Dropdown disabled={isFileDisabled}>
+                            File
+                            <ul>
+                                <li data-ukt-item>New</li>
+                            </ul>
+                        </Dropdown>
+                        <Dropdown>
+                            Edit
+                            <ul>
+                                <li data-ukt-item>Undo</li>
+                            </ul>
+                        </Dropdown>
+                    </>
+                );
+            };
+            render(
+                <Menubar>
+                    <Members />
+                </Menubar>,
+            );
+
+            expect(
+                screen.getByRole('menuitem', { name: 'File' }).getAttribute('tabindex'),
+            ).toBe('0');
+
+            await user.click(screen.getByRole('button', { name: 'disable File' }));
+
+            expect(
+                screen.getByRole('menuitem', { name: 'Edit' }).getAttribute('tabindex'),
+            ).toBe('0');
+            expect(
+                screen.getByRole('menuitem', { name: 'File' }).getAttribute('tabindex'),
+            ).toBe('-1');
+        });
     });
 
     it('keeps at most one menu open at a time', async () => {
--- a/packages/dropdown/src/Menubar.test.tsx
+++ b/packages/dropdown/src/Menubar.test.tsx
@@ -1,7 +1,7 @@
 // @vitest-environment happy-dom
-import { act, cleanup, fireEvent, render, screen } from '@testing-library/react';
+import { act, cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { userEvent } from '@testing-library/user-event';
-import { useState } from 'react';
+import { Activity, useState } from 'react';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
 import Dropdown, { Menubar } from './Dropdown.js';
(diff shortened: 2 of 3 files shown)


packages/dropdown/src/Menubar.tsx
```diff
@@ -145,9 +145,14 @@ export default function Menubar({ children, className, style }: MenubarProps) {
...
-                    // up a tab stop stranded on an unmounted member
-                    reconcile();
+                    // Say the set has settled, so the effect above can pick up
+                    // a tab stop stranded on an unmounted member. Members
+                    // re-register on every render, so this cleanup runs far
+                    // more often than one actually leaves; only the holder
+                    // going (or becoming ineligible, which re-renders it here
+                    // too) can strand the tab stop, so bumping for the rest
+                    // would schedule a Menubar render that changes nothing.
+                    if (member.element === tabbableElement) reconcile();
```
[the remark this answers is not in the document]
acusti: Valid, and confirmed by measurement — thanks. Fixed in a6ae04c.

I instrumented the bar with a `<Profiler>` and ran three open/close cycles on the tab stop holder's own menu, which re-registers it repeatedly while it stays eligible. On the compiled build:

| | Menubar commits |
|---|---|
| unconditional bump (before this PR) | 19 |
| `member.element === tabbableElement` (b676aa7) | **19** |
| after a6ae04c | 9 |

So the first narrowing bought exactly nothing in that case, as you said — the 27 → 12 figure in the PR description came only from *other* members churning during hover-switching.

The fix splits the two situations that were conflated:

- **Genuine departure** — React has already detached the member's element by the time the passive-effect cleanup runs, so `member.element === tabbableElement && !member.element.isConnected` distinguishes a real leave from a re-registration.
- **Eligibility transition** — moved to the registration side, where the freshly registered member reflects the props that changed: a holder that re-registers unable to hold the tab stop returns `null` and the effect rehomes it. Every other re-registration returns `current` untouched, so React bails out rather than re-rendering the bar.

`isConnected` alone would have been wrong — it misses a holder that becomes `disabled` without unmounting — which is why the eligibility half moved to registration rather than being dropped.

---

packages/dropdown/src/Menubar.test.tsx
```diff
@@ -327,6 +327,54 @@ describe('@acusti/dropdown Menubar', () => {
                 screen.getByRole('menuitem', { name: 'Edit' }).getAttribute('tabindex'),
             ).toBe('0');
         });
+
+        it('hands the tab stop on when the holder becomes disabled', async () => {
```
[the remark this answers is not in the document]
acusti: You're right that this test passes pre-fix — that's deliberate, but your underlying point stands and it led somewhere useful.

**What the test is for.** It guards the *narrowing*, not the render saving. The pre-fix code is correct-but-wasteful, so no correctness test can fail on it; the risk this change introduces is narrowing too far and stranding the tab stop. Concretely, `if (!member.element.isConnected) reconcile()` — a plausible-looking narrowing — passes all 33 pre-existing tests and still strands the tab stop when the holder becomes `disabled` without unmounting. This test fails on that. So it isn't redundant coverage, it's a guard against a specific wrong fix, and it earned its place immediately: the eligibility half of a6ae04c is exactly what it forced.

**On asserting the render count — I tried it and removed it.** I added a `<Profiler>`-based test asserting commits for three open/close cycles. It works, but it can't be a reliable guard in this repo:

| config | pre-fix | post-fix |
|---|---|---|
| `packages/dropdown/vite.config.js` (`react: true`, compiler on) | 19 | 9 |
| root `vitest.config.js` (no react plugin, compiler off) | 19 | **18** |

CI runs `bun run test` from the root, which has no react plugin, so the compiler is off in the run that gates this PR — and there the difference is 19 vs 18. Any threshold separating those is one commit wide and would flake on a React or Testing Library bump. Without the compiler the bar re-renders for other reasons anyway, which masks the reconcile bumps; with it, they're the dominant remaining cost, which is why the shipped build is where the win shows up.

So I've left the perf change guarded by measurement recorded in the commit message and changeset rather than by a brittle assertion, and kept the correctness tests. Flagging the gap plainly: nothing in CI will catch a future regression back to unconditional bumping. If you'd want that covered, the honest way is a dedicated compiler-enabled test project rather than a threshold in the shared suite — happy to open a follow-up issue if you think it's worth it.

---

packages/dropdown/src/Menubar.tsx
```diff
@@ -140,14 +140,42 @@ export default function Menubar({ children, className, style }: MenubarProps) {
...
+                    // more often than one actually leaves: only the holder
+                    // actually going strands the tab stop, and React has
+                    // already detached its element by the time this runs,
+                    // which is what tells a departure from a re-render.
+                    // Bumping for the rest would schedule a Menubar render
+                    // that changes nothing. Losing eligibility without leaving
+                    // is handled on the registration side above.
+                    if (
+                        member.element === tabbableElement &&
+                        !member.element.isConnected
```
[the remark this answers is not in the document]
acusti: Good catch — valid, and reproduced before fixing. Fixed in 8ce1ca3.

`Activity` is real here (React 19.2.8 installed, and this package's peer range is `^19`), so I wrote the case rather than reasoning about it. With the `isConnected` check, hiding the tab stop holder with `<Activity mode="hidden">` gave exactly what you described:

```
fileStillConnected: true
editTabindex: "-1"   // stop stranded on the hidden trigger
```

You're also right about the underlying mistake: `isConnected` was answering "what does the DOM look like?" when the real question is "did this registration go away?". The unconditional bump this PR replaced handled `Activity` correctly, so my check was a regression, not just a missed case.

The fix stops inspecting the DOM entirely. A member that merely re-rendered re-registers during the same commit, so the decision is deferred until the commit has settled and the bump fires only if that element never came back:

```ts
if (member.element !== tabbableElement) return;
queueMicrotask(() => {
    for (const registered of membersRef.current) {
        if (registered.element === member.element) return;
    }
    reconcile();
});
```

That covers an unmount, a hidden subtree, and any other way effects are torn down, without a special case for each. The holder-churn measurement is unchanged at 9 commits, so nothing was traded away for it.

Added as a test (`hands the tab stop on when the holder is hidden with <Activity>`) which fails on the `isConnected` version and passes on this one — so this specific regression can't come back silently.

---

packages/dropdown/src/Menubar.tsx
```diff
@@ -140,14 +140,46 @@ export default function Menubar({ children, className, style }: MenubarProps) {
...
+                    // runs far more often than one actually leaves, and
+                    // bumping for the rest would schedule a Menubar render
+                    // that changes nothing. Only the holder can strand the
+                    // stop, and a re-render re-registers it during this same
+                    // commit — so let the commit settle and bump only if this
+                    // element never came back. Whether the element is still in
+                    // the document doesn’t answer it: <Activity mode="hidden">
+                    // unmounts effects while keeping the DOM in place. Losing
+                    // eligibility without leaving is handled on the
+                    // registration side above.
```
acusti: can we say this with less words? my eyes glaze over when i see a wall of text like this

packages/dropdown/src/Menubar.tsx
```diff
@@ -140,14 +140,46 @@ export default function Menubar({ children, className, style }: MenubarProps) {
...
+                    // gets it
+                    if (current == null) {
+                        return canHoldTabStop(member) ? member.element : current;
+                    }
+                    // A holder that re-registers no longer able to hold it
+                    // (disabled, or no longer a menu) gives it up here, where
+                    // the fresh member reflects the props that changed; the
+                    // effect above then rehomes it. Every other
+                    // re-registration returns `current` untouched, so React
+                    // bails out instead of re-rendering the bar.
```
acusti: > A holder that re-registers no longer able to hold it (disabled, or no longer a menu) gives it up here

i don't really get what that means. also, this is not so bad as the one below, but it's nice when a reader doesn't have to get through so many lines of text

acusti: Fair — that sentence was doing too much at once. Rewritten in 7c12fd9:

```ts
// the holder re-registering ineligible — now disabled, or
// no longer a menu — drops the stop for the effect above
// to rehome. Every other re-registration returns `current`
// unchanged, so React bails out of rendering the bar.
```

The bit that wasn't landing: "gives it up **here**" was trying to say *this is the place the change is noticed*, because members re-register on every render and the fresh `member` closes over the new props — so a `disabled` that just flipped shows up in `canHoldTabStop(member)` here, and nowhere else. That's now just "drops the stop for the effect above to rehome", with the where-and-why left to the code.

Also removed a duplicate: the "claim the tab stop when it's going spare" comment had been left in twice, once above `setTabbableElement` and once inside it.

---

packages/dropdown/src/Menubar.tsx
acusti: Agreed, that one had grown to fourteen lines. Cut to seven in 7c12fd9:

```ts
// Only the holder leaving strands the stop, and members
// re-register on every render — so wait out the commit and
// bump only if this element never came back. Still being in
// the document proves nothing: <Activity mode="hidden">
// tears down effects but keeps the DOM. Releasing the stop
// here instead would hand it to whichever member
// re-registers first, not back to the holder.
```

What went and why: the "bumping for the rest would schedule a Menubar render that changes nothing" line was restating the perf rationale already in the commit message and changeset, and the separate four-line "the tab stop is deliberately not released here" block below folded into the last sentence. What stayed is the three things a reader can't get from the code — why only the holder matters, why the check has to wait for the commit, and why `isConnected` would be the wrong test.

Whole hunk is 28 lines lighter.

---

.changeset/dropdown-menubar-reconcile-only-for-holder.md
```diff
@@ -0,0 +1,21 @@
...
+
+Only the holder really going can strand the tab stop, and a member that
+merely re-rendered re-registers during the same commit — so the bump now
+waits for the commit to settle and fires only if that member never came
+back. Losing eligibility without leaving (a holder that becomes disabled,
+or stops being a menu) is settled where the member re-registers, against
+the props that changed.
+
+Measured on the compiled build, opening and closing the tab stop holder’s
+own menu three times: 19 Menubar commits before, 9 after.
```
acusti: this should be short, maybe even just the top line. this is a change of minimal interest to package consumers that would often be summarized as “Performance improvements”, but obviously we can do better than that. but we shouldn’t explain the previous behavior, nor should we explain the internals, just an approachable/accessible/brief explanation of the change

packages/dropdown/src/Menubar.tsx
```diff
@@ -138,20 +138,37 @@ export default function Menubar({ children, className, style }: MenubarProps) {
...
             registerMember(member: MenubarMember) {
                 membersRef.current.add(member);
-                // Claim the tab stop when it’s going spare; members register
-                // in document order, so the first enabled one gets it
-                setTabbableElement((current) =>
-                    current == null && canHoldTabStop(member) ? member.element : current,
-                );
+                setTabbableElement((current) => {
+                    // members register in document order, so a stop going
+                    // spare lands on the first eligible one
```
acusti: hadn’t noticed this before, but i don’t know what “a stop going spare” means

packages/dropdown/src/Menubar.tsx
```diff
@@ -138,20 +138,37 @@ export default function Menubar({ children, className, style }: MenubarProps) {
...
+                    if (current == null) {
+                        return canHoldTabStop(member) ? member.element : current;
+                    }
+                    // the holder re-registering ineligible — now disabled, or
+                    // no longer a menu — drops the stop for the effect above
+                    // to rehome. Every other re-registration returns `current`
```
acusti: i don't understand “the holder re-registering ineligible drops the stop” as a phrase. i can’t tell what the subject and verb are.

packages/dropdown/src/Menubar.tsx
acusti: Yeah, that's British-idiom filler doing no work. "Going spare" meant "unclaimed", which the sentence never actually says. Fixed in 66f2ca1:

```ts
// no member holds the tab stop yet, so the first
// eligible one to register takes it — and they register
// in document order
```

---

packages/dropdown/src/Menubar.tsx
acusti: You're right that it doesn't parse. I'd compressed "the holder, re-registering while ineligible, drops the stop" into a noun pile where "re-registering ineligible" reads as neither a clause nor a modifier, and "the holder" is doing double duty as subject of two different verbs. Fixed in 66f2ca1:

```ts
// this member holds the tab stop but can no longer keep
// it: it has just been disabled, or has stopped being a
// menu. Clearing the stop lets the effect above hand it
// to another member. Every other re-registration returns
// `current` untouched, so React skips re-rendering the bar.
```

Now each sentence has a plain subject and verb: *this member* holds/can't keep, *clearing the stop* lets the effect hand it on.

---

.changeset/dropdown-menubar-reconcile-only-for-holder.md
acusti: Agreed — I'd written it for a reviewer of this PR, not for someone reading the changelog. Cut to the summary line alone in 66f2ca1:

```md
---
'@acusti/dropdown': patch
---

Halve the renders a Menubar does as its menus open and close
```

The two paragraphs on what the reducer used to do and how the fix works were both things a consumer can't act on, and the 19→9 measurement is only meaningful next to that explanation — all of it lives in the commit messages and the PR instead.

---

.changeset/dropdown-menubar-reconcile-only-for-holder.md
```diff
@@ -0,0 +1,21 @@
...
+early-returned because the tab stop was still held by an eligible member.
+
+Only the holder really going can strand the tab stop, and a member that
+merely re-rendered re-registers during the same commit — so the bump now
+waits for the commit to settle and fires only if that member never came
+back. Losing eligibility without leaving (a holder that becomes disabled,
+or stops being a menu) is settled where the member re-registers, against
+the props that changed.
+
+Measured on the compiled build, opening and closing the tab stop holder’s
```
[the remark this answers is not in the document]
acusti: Valid, and already fixed in 66f2ca1 — this review ran against 7c12fd9, the previous head. acusti raised the same point, and the changeset is now the summary line alone:

```md
---
'@acusti/dropdown': patch
---

Halve the renders a Menubar does as its menus open and close
```

The measurement went further than you suggested: rather than moving to the PR description, it was already there, so it's simply gone from the changeset along with the two paragraphs of mechanism.

---

PR #436, opened 2026-08-30
Forward aria-* attributes and role in InputText and CSSValueInput

acusti:
Neither component’s `Props` included any of the ARIA attributes, so an `aria-label` or `aria-describedby` was both a type error and silently dropped. `InputText` could only be named by a wrapping `<label>`, and it couldn’t take the combobox semantics `@acusti/dropdown` injects when it clones a custom trigger (`role`, `aria-expanded`, `aria-controls`, `aria-autocomplete`). `CSSValueInput` had no way to annotate its input at all, and its only accessible name came from `label` or the `aria-label` it derives from `title` for the icon-only case — neither overridable by the consumer.

Both `Props` types now intersect React’s `AriaAttributes` and add `role`, and whatever the component doesn’t destructure is spread onto the element.

### Approach

Intersecting `AriaAttributes` rather than hand-enumerating the attributes or extending `InputHTMLAttributes`:

- Hand-enumerating means ~50 attributes to write out and keep in sync with the ARIA spec, for nothing React’s type doesn’t already give.
- An open spread over `InputHTMLAttributes` would let a consumer pass `value`, `defaultValue`, `onInput`, or `onSelect` and quietly break `InputText`’s semi-controlled behavior — `defaultValue` fights `initialValue`, `onInput` clobbers the multiLine resize handler, `onSelect` kills `selectTextOnFocus`. The curated prop list is deliberate.

`AriaAttributes` gets both: nothing to maintain, and the prop type stays closed — props the components don’t model remain type errors instead of passing through.

`role` is included alongside the aria attributes because `Dropdown` injects it on a cloned trigger (`Dropdown.tsx:1453`); without it, an `InputText` trigger would get `aria-expanded` and not the role that makes it valid on a text input.

### Placement

- **`InputText`** — the spread comes first in the JSX, so the props the component owns (`readOnly` under `doubleClickToEdit`, `defaultValue`, its wrapped handlers) always win. No collision exists today, since the rest is only `aria-*` and `role`, but a test pins the invariant.
- **`CSSValueInput`** — the spread lands on the nested `InputText`, not the wrapping `<label>`, so the annotations reach the element carrying the `textbox` role. The wrapper’s existing `aria-label={label ? undefined : title}` is left alone: accessible-name precedence already puts a consumer’s `aria-label` on the input ahead of any name derived from the wrapping label, and a test covers that (`aria-label="Width"` wins over both `label="W"` and `title="Element width"`).

### Testing

Five tests added — forwarding to `input` and to `textarea` under `multiLine`, the ownership guard, forwarding to the input rather than the wrapper, and the naming precedence. Full repo: 395 tests pass, `lint` / `tsc` / `format:check` clean.

READMEs updated and a minor changeset added for each package.

https://claude.ai/code/session_01X2mNX6d9jrshUKLugpsuWz

---
_Generated by [Claude Code](https://claude.ai/code/session_01X2mNX6d9jrshUKLugpsuWz)_

<a href="https://cubic.dev/pr/acusti/uikit/pull/436?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/css-value-input-aria-props.md
+++ b/.changeset/css-value-input-aria-props.md
@@ -1,0 +1,17 @@
+---
+'@acusti/css-value-input': minor
+---
+
+Forward aria-\* attributes and role to the underlying input
+
+`Props` didn’t include any of the ARIA attributes, so annotating the input
+— `aria-describedby` for a hint, `aria-invalid` on a rejected value —
+wasn’t possible. The only name the input could get came from `label` or,
+for the icon-only case, the `aria-label` the component puts on its wrapping
+`<label>` from `title`, neither of which the consumer could override.
+
+`Props` now intersects React’s `AriaAttributes` and adds `role`, and any of
+those props not consumed by the component are spread onto the nested
+`InputText` rather than the wrapping `<label>`, so they annotate the input
+itself. An `aria-label` passed in names the input and, by the accessible
+name precedence, wins over the wrapper’s label text or title.
--- a/.changeset/css-value-input-aria-props.md
+++ b/.changeset/css-value-input-aria-props.md
@@ -11,7 +11,6 @@ for the icon-only case, the `aria-label` the component puts on its wrapping
 `<label>` from `title`, neither of which the consumer could override.
 
 `Props` now intersects React’s `AriaAttributes` and adds `role`, and any of
-those props not consumed by the component are spread onto the nested
-`InputText` rather than the wrapping `<label>`, so they annotate the input
-itself. An `aria-label` passed in names the input and, by the accessible
-name precedence, wins over the wrapper’s label text or title.
+those props not consumed by the component are spread onto the nested input.
+An `aria-label` passed in names the input and, by the accessible name
+precedence, wins over the wrapper’s label text or title.
--- a/.changeset/input-text-aria-props.md
+++ b/.changeset/input-text-aria-props.md
@@ -13,7 +13,4 @@ semantics it injects.
 
 `Props` now intersects React’s `AriaAttributes` and adds `role`, and any of
 those props not consumed by the component are spread onto the `<input>` or
-`<textarea>`. The spread comes first, so the props the component owns
-(`readOnly` under `doubleClickToEdit`, `defaultValue`, its event handlers)
-still win, and the prop type stays closed: props `InputText` doesn’t model
-remain type errors rather than passing through.
+`<textarea>`.
--- a/.changeset/input-text-aria-props.md
+++ b/.changeset/input-text-aria-props.md
@@ -1,0 +1,19 @@
+---
+'@acusti/input-text': minor
+---
+
+Forward aria-\* attributes and role to the underlying input
+
+`Props` didn’t include any of the ARIA attributes, so an `aria-label` or an
+`aria-describedby` passed to `InputText` was both a type error and silently
+dropped — the only accessible name available was one supplied by a wrapping
+`<label>`. It also meant a parent that annotates its trigger by cloning it,
+like `@acusti/dropdown`, couldn’t give an `InputText` trigger the combobox
+semantics it injects.
+
+`Props` now intersects React’s `AriaAttributes` and adds `role`, and any of
+those props not consumed by the component are spread onto the `<input>` or
+`<textarea>`. The spread comes first, so the props the component owns
+(`readOnly` under `doubleClickToEdit`, `defaultValue`, its event handlers)
+still win, and the prop type stays closed: props `InputText` doesn’t model
+remain type errors rather than passing through.
--- a/packages/css-value-input/src/CSSValueInput.tsx
+++ b/packages/css-value-input/src/CSSValueInput.tsx
@@ -10,6 +10,8 @@ import {
 import InputText, { type InputElement } from '@acusti/input-text';
 import clsx from 'clsx';
 import {
+    type AriaAttributes,
+    type AriaRole,
     type ChangeEvent,
     type FocusEvent,
     type KeyboardEvent,
@@ -20,7 +22,13 @@ import {
     useRef,
 } from 'react';
 
-export type Props = {
+/**
+ * Every aria-* attribute (and role) is forwarded as-is to the underlying
+ * input, not to the wrapping <label>, so the input is annotated directly. An
+ * aria-label passed here names the input and takes precedence over the name
+ * it would otherwise get from props.label or props.title.
+ */
+export type Props = AriaAttributes & {
     /**
      * Boolean indicating if the user can submit an empty value (i.e. clear
      * the value). Defaults to true.
@@ -52,6 +60,7 @@ export type Props = {
     onSubmitValue: (value: string) => unknown;
     placeholder?: string;
     ref?: Ref<HTMLInputElement>;
+    role?: AriaRole;
     step?: number;
     tabIndex?: number;
     title?: string;
@@ -88,6 +97,7 @@ export default function CSSValueInput({
     unit = DEFAULT_UNIT_BY_CSS_VALUE_TYPE[cssValueType],
     validator,
     value: valueFromProps,
+    ...restProps
 }: Props) {
     // props.value should be a string; if it’s a number, convert it here
     const value =
@@ -262,6 +272,7 @@ export default function CSSValueInput({
             ) : null}
             <div className={`${ROOT_CLASS_NAME}-value`}>
                 <InputText
+                    {...restProps}
                     disabled={disabled}
                     discardOnEscape
                     initialValue={value}
--- a/packages/css-value-input/src/CSSValueInput.tsx
+++ b/packages/css-value-input/src/CSSValueInput.tsx
@@ -24,9 +24,8 @@ import {
 
 /**
  * Every aria-* attribute (and role) is forwarded as-is to the underlying
- * input, not to the wrapping <label>, so the input is annotated directly. An
- * aria-label passed here names the input and takes precedence over the name
- * it would otherwise get from props.label or props.title.
+ * input. An aria-label passed here names the input and takes precedence over
+ * the name it would otherwise get from props.label or props.title.
  */
 export type Props = AriaAttributes & {
     /**
--- a/packages/docs/stories/InputText.stories.tsx
+++ b/packages/docs/stories/InputText.stories.tsx
@@ -87,8 +87,6 @@ const SUBMIT_ON_ENTER_PROPS = {
     submitOnEnter: true,
 };
 
-const MIN_PASSPHRASE_LENGTH = 8;
-
 const formatDate = new Intl.DateTimeFormat(undefined, {
     timeStyle: 'medium',
 }).format;
@@ -410,6 +408,8 @@ export const InputWithAriaLabel: Story = {
         },
     },
 };
+
+const MIN_PASSPHRASE_LENGTH = 8;
 
 export const InputWithHintAndValidationState: Story = {
     args: {
--- a/packages/docs/stories/InputText.stories.tsx
+++ b/packages/docs/stories/InputText.stories.tsx
@@ -36,6 +36,7 @@ type Story = StoryObj<typeof InputText>;
 
 export const EmptyInput: Story = {
     args: {
+        'aria-label': 'Text',
         className: 'input-text',
         name: 'empty',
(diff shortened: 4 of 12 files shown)


packages/input-text/src/InputText.test.tsx
```diff
@@ -409,4 +409,38 @@ describe('CSSValueInput.tsx', () => {
...
+
+    it('forwards aria-* attributes to the textarea when multiLine', () => {
+        render(<InputText aria-describedby="hint-id" aria-label="Notes" multiLine />);
+        const textarea = screen.getByRole('textbox');
+        expect(textarea.tagName).toBe('TEXTAREA');
+        expect(textarea.getAttribute('aria-label')).toBe('Notes');
+        expect(textarea.getAttribute('aria-describedby')).toBe('hint-id');
+    });
+
+    it('doesn’t let forwarded props override the props the component owns', () => {
```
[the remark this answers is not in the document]
acusti: Valid, and confirmed empirically: moving `{...restProps}` to the end of the JSX left all 19 tests green, so the guard was vacuous exactly as described.

Rather than rename it, I made it able to fail (3936a88). The reasoning about `Props` is right — every prop it models is destructured, so no modeled prop can ever collide in the rest — but that means the collision only exists for a prop `Props` doesn't model, which an untyped consumer can still pass. `onDoubleClick` is one: the component sets it from `doubleClickToEdit` and never destructures it. The test now passes one in and asserts the component's handler still wins:

```tsx
const onDoubleClick = vi.fn<() => void>();
render(<InputText {...({ onDoubleClick } as Props)} doubleClickToEdit />);
expect(input.readOnly).toBe(true);
await user.dblClick(input);
expect(input.readOnly).toBe(false);
expect(onDoubleClick).not.toHaveBeenCalled();
```

Verified in both directions: it fails (`expected true to be false`) with the spread moved last, and passes with it first.

One note for anyone reading the suggested rename — a first replacement I tried used `defaultValue`, which is also vacuous: the mount effect (`inputRef.current.value = valueFromProps`) overwrites the DOM value regardless of spread order. `onDoubleClick` avoids that.

---

packages/input-text/src/InputText.test.tsx
```diff
@@ -409,4 +409,38 @@ describe('CSSValueInput.tsx', () => {
         expect(height).toBeGreaterThanOrEqual(50);
         expect(height).toBeLessThanOrEqual(100);
     });
+
+    it('forwards aria-* attributes and role to the input', () => {
```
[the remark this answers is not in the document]
acusti: Valid — and it predates this PR: the `describe` in `InputText.test.tsx` has been labelled `CSSValueInput.tsx` since the file was created, so the entire suite (not just the new tests) reported under the wrong component. Every other test file in the repo names its own subject.

Renamed the existing block to `describe('InputText.tsx')` rather than nesting a second one, since the whole file tests `InputText` — 839a19d.

---

.changeset/input-text-aria-props.md
```diff
@@ -0,0 +1,19 @@
...
+`<label>`. It also meant a parent that annotates its trigger by cloning it,
+like `@acusti/dropdown`, couldn’t give an `InputText` trigger the combobox
+semantics it injects.
+
+`Props` now intersects React’s `AriaAttributes` and adds `role`, and any of
+those props not consumed by the component are spread onto the `<input>` or
+`<textarea>`. The spread comes first, so the props the component owns
+(`readOnly` under `doubleClickToEdit`, `defaultValue`, its event handlers)
+still win, and the prop type stays closed: props `InputText` doesn’t model
+remain type errors rather than passing through.
```
acusti: this is getting deeper into the internals of the component than is necessary for a changelog 
```suggestion
`<textarea>`.
```

.changeset/css-value-input-aria-props.md
```diff
@@ -0,0 +1,17 @@
...
+`Props` didn’t include any of the ARIA attributes, so annotating the input
+— `aria-describedby` for a hint, `aria-invalid` on a rejected value —
+wasn’t possible. The only name the input could get came from `label` or,
+for the icon-only case, the `aria-label` the component puts on its wrapping
+`<label>` from `title`, neither of which the consumer could override.
+
+`Props` now intersects React’s `AriaAttributes` and adds `role`, and any of
+those props not consumed by the component are spread onto the nested
+`InputText` rather than the wrapping `<label>`, so they annotate the input
+itself. An `aria-label` passed in names the input and, by the accessible
```
acusti: we can simplify:

```suggestion
those props not consumed by the component are spread onto the nested input. An `aria-label` passed in names the input and, by the accessible
```

packages/css-value-input/src/CSSValueInput.tsx
```diff
@@ -20,7 +22,13 @@ import {
     useRef,
 } from 'react';
 
-export type Props = {
+/**
+ * Every aria-* attribute (and role) is forwarded as-is to the underlying
+ * input, not to the wrapping <label>, so the input is annotated directly. An
+ * aria-label passed here names the input and takes precedence over the name
+ * it would otherwise get from props.label or props.title.
```
acusti:
```suggestion
 * Every aria-* attribute (and role) is forwarded as-is to the underlying
 * input. An
 * aria-label passed here names the input and takes precedence over the name
 * it would otherwise get from props.label or props.title.
```

packages/input-text/README.md
```diff
@@ -79,8 +79,16 @@ function SelectOnFocusInput() {
 
 ### InputText Component
 
+Every `aria-*` attribute and `role` is also accepted and forwarded as-is to
+the underlying `<input>`/`<textarea>`, so the component can be labelled and
+annotated exactly like the native element:
```
acusti:
```suggestion
Every `aria-*` attribute and `role` is also accepted and forwarded as-is to
the underlying `<input>`/`<textarea>`:
```

packages/input-text/src/InputText.test.tsx
```diff
@@ -19,7 +21,7 @@ vi.unstubAllGlobals();
 
 afterEach(cleanup);
 
-describe('CSSValueInput.tsx', () => {
+describe('InputText.tsx', () => {
```
acusti, line 24: good catch!

packages/input-text/src/InputText.tsx
```diff
@@ -14,7 +16,13 @@ import {
 
 export type InputElement = HTMLInputElement | HTMLTextAreaElement;
 
-export type Props = {
+/**
+ * Every aria-* attribute (and role) is forwarded as-is to the underlying
+ * <input>/<textarea>, so the component can be labelled and annotated the same
+ * way the native element is — including by a parent that clones it to inject
```
acusti:
```suggestion
 * <input>/<textarea>, so the component can be labelled and annotated, including by a parent that clones it to inject
```

acusti (commented):
looks good. for the comment suggestions, i didn't bother re-wrapping the text, i'll leave that for you to figure out.

the other thing i want in this PR is to update the stories where appropriate to demonstrate the a11y best practices for how to use this component.

.changeset/input-text-aria-props.md
acusti: Agreed — cut in 5703119. The changeset now stops at "spread onto the `<input>` or `<textarea>`"; the spread order and the closed prop type are implementation detail that belongs in the PR, not the changelog.

---

.changeset/css-value-input-aria-props.md
acusti: Applied in 5703119, re-wrapped:

```
`Props` now intersects React’s `AriaAttributes` and adds `role`, and any of
those props not consumed by the component are spread onto the nested input.
An `aria-label` passed in names the input and, by the accessible name
precedence, wins over the wrapper’s label text or title.
```

---

packages/css-value-input/src/CSSValueInput.tsx
acusti: Applied in 5703119, re-wrapped:

```
/**
 * Every aria-* attribute (and role) is forwarded as-is to the underlying
 * input. An aria-label passed here names the input and takes precedence over
 * the name it would otherwise get from props.label or props.title.
 */
```

---

packages/input-text/README.md
acusti: Applied in 5703119 — the trailing clause was restating what the sentence already says, and the example below it shows the point anyway.

---

packages/input-text/src/InputText.tsx
acusti: Applied in 5703119, re-wrapped:

```
/**
 * Every aria-* attribute (and role) is forwarded as-is to the underlying
 * <input>/<textarea>, so the component can be labelled and annotated,
 * including by a parent that clones it to inject combobox semantics, e.g.
 * @acusti/dropdown’s custom trigger.
 */
```

---

packages/docs/stories/InputText.stories.tsx
```diff
@@ -82,6 +87,8 @@ const SUBMIT_ON_ENTER_PROPS = {
     submitOnEnter: true,
 };
 
+const MIN_PASSPHRASE_LENGTH = 8;
```
acusti: can we move this down directly above the story that uses it?

acusti: Moved in a701b77 — it now sits directly above `InputWithHintAndValidationState`, its only user. Grouping it with the shared story constants put ~250 lines between the two and implied the stories in between might use it.

---

packages/input-text/src/InputText.tsx
```diff
@@ -75,6 +83,7 @@ export type Props = {
     readOnly?: boolean;
     ref?: Ref<HTMLInputElement>;
     required?: boolean;
+    role?: AriaRole;
```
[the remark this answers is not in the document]
acusti: The finding is correct — under `multiLine` the ref receives an `HTMLTextAreaElement` while `Props.ref` promises `HTMLInputElement`. But it predates this PR: `ref` isn't touched here, and nothing about the aria forwarding affects it. Flagging rather than folding it in, since it's a public type change on a released package and unrelated to what this PR is for.

I checked the two things that decide whether it's a safe fix:

**Widening isn't a breaking change.** My first worry was callback refs — under `strictFunctionTypes` a consumer's `(el: HTMLInputElement | null) => void` should stop being assignable to a widened `Ref<HTMLInputElement | HTMLTextAreaElement>`. It doesn't, because React types `RefCallback` with the bivariance hack. Verified by compiling both a narrow object ref and a narrow callback ref against the widened signature under `--strict`: no errors.

**The patch is two lines and comes back clean:**

```diff
-    ref?: Ref<HTMLInputElement>;
+    ref?: Ref<InputElement>;

-type InputRef = HTMLInputElement | null;
+type InputRef = InputElement | null;
```

`InputElement` is already exported from this module. Applied locally, `bun run tsc` passes across the monorepo and all 19 `input-text` tests pass. `CSSValueInput` needs no change — it never sets `multiLine`, so its `Ref<HTMLInputElement>` is accurate, and it still passes through to the widened prop.

The discriminated-union alternative would be more precise, but it makes `Props` conditional on `multiLine` and complicates every consumer that holds a ref without knowing the mode — not worth it over the widening.

The practical risk is low, for what it's worth: internally the component only touches `.style`, `.value`, and `.scrollHeight`, which both elements have. The exposure is a consumer reaching for an input-only API on a ref from a multi-line instance.

@acusti — happy to push this here with a patch changeset, or open it as its own PR. Say which and I'll do it; leaving the thread open until then.

---

acusti:
not going to address the ref type in this PR, it's preexisting and untouched by these changes, but i will open a fresh branch to address it

packages/input-text/src/InputText.tsx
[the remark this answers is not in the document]
acusti: Confirmed by @acusti in the PR thread: out of scope here, to be handled on a fresh branch. Nothing in this PR touches `ref` or `InputRef`, so resolving. The two-line patch above still applies as-is against `main` if it's useful there.

---

PR #448, opened 2026-09-08
feat(vite-plugin-svg-react): add an opt-in OXVG optimization pass

### LINKED ISSUE #422: vite-plugin-svg-react: optional OXVG optimization pass
Since #421, `@acusti/vite-plugin-svg-react` generates component modules directly — no svgr, no Babel, zero runtime dependencies — and deliberately leaves SVG optimization out of scope (svgr offered it via `@svgr/plugin-svgo`).

An optional pre-optimization pass built on [OXVG](https://github.com/noahbald/oxvg) — the Rust, SVGO-compatible SVG toolchain — would fit the plugin’s all-native oxc/rolldown pipeline and could slot in ahead of component generation without reintroducing a JS compiler. Sketch:

- opt-in via a plugin option (e.g. `optimize: true` or an OXVG config passthrough), default off
- runs on the raw SVG source before parsing/emission, so the rest of the pipeline is untouched
- OXVG stays an optional peer/opt-in dependency so the default install remains dependency-free

If you want this, add your 👍 to this issue.

acusti:
Closes #422.

Adds an `optimize` option to `@acusti/vite-plugin-svg-react` that runs each SVG through [OXVG](https://github.com/noahbald/oxvg) — the Rust, SVGO-compatible toolchain — on the raw source, ahead of parsing and emission. Nothing reintroduces a JS compiler. Off by default.

```ts
svgReact({ optimize: true });
```

`@oxvg/napi` is an **optional peer dependency**, so the default install stays dependency-free. Using `optimize` means installing it yourself (`npm i -D @oxvg/napi`) — `peerDependenciesMeta.optional` tells package managers not to auto-install. That's deliberate: the package is 0.1 MB but pulls a 12.5–16 MB native binary per platform, which is a lot to impose on every install of a plugin whose pitch is "zero dependencies", for an option that's off by default.

## `optimize: true` is the default preset minus `cleanupIds`

Nothing it runs renames an id or a class.

`cleanupIds` minifies ids and removes unreferenced ones, which is a rename the optimizer can't verify: anything pointing at an id from outside the file — app CSS, `getElementById`, an `aria-labelledby` — breaks silently. It's also what makes inlined components collide, since every file's ids collapse to the same `a` and `b`, and then an internal `<use href="#a">` resolves against whichever component rendered first. `@svgr/plugin-svgo` works around that by adding `prefixIds` on top, which renames the ids a second time and renames class names with them; leaving them alone keeps ids exactly as unique — or not — as the author made them.

That's a default, not a verdict: the README shows `extend({ type: 'Default' })` for opting back into the renames. Passing an object instead of `true` hands it to OXVG's `optimise` as-is — an OXVG config is the complete list of optimizations to run, not overrides on top of a preset.

## What `optimize` costs

**It narrows what builds.** OXVG parses before this plugin's deliberately tolerant parser, and rejects three things `parseSVG` accepts on purpose: a doctype whose internal subset declares entities the root element references (classic Illustrator 10–CS4 wrote its Adobe namespaces that way), an unknown entity name like `&nbsp;`, and a minimized attribute (`<svg hidden>`). Each fails by name and each still builds with `optimize` off. Documented in the README under **What `optimize` stops accepting**, and pinned by tests.

**OXVG rejects any document with a DTD**, so the prolog is blanked out — spaces, newlines kept — rather than cut off, which keeps the `line:column` in OXVG's errors pointing at the line the file actually has. Cutting it reported line 1 for what the editor calls line 5.

**It ignores job names it doesn't recognize**, silently: `optimise(svg, { notAJob: {} })` doesn't throw. There's no drift-free way to validate them at runtime — the presets expose only 36 of 57 job keys, and `extend` is stricter about params than `optimise`, so a round-trip check would reject valid configs. The README shows `satisfies Jobs` at the call site for consumers who want the names typed.

## Failure modes

Both resolve when Vite resolves the config, so the message names the option rather than whichever SVG loaded first:

- missing `@oxvg/napi` → `the optimize option needs @oxvg/napi, which failed to load. Install it (…) or remove the optimize option.` (verified by hiding the package from the store). Not memoized on failure, so installing it and re-saving `vite.config` recovers in the same process.
- a config OXVG rejects → `OXVG rejected the optimize config: …`

## Notes on the types

`OptimizeConfig` is `object`, not `Record<string, unknown>`: OXVG's `Jobs` is an interface, and TypeScript gives interfaces no implicit index signature, so a Record rejects the config `extend` returns. The published `.d.ts` still carries no `@oxvg/napi` reference. A test builds its config with `extend` and passes it to the plugin, so `bun run tsc` catches a regression there.

## Commits

1. `chore: sync bun.lock with vite-plugin-react-compiler 0.5.0` — pre-existing drift on `main` that any `bun install` rewrites, kept out of the feature's blame.
2. `refactor(vite-plugin-svg-react): extract the prolog scan from parseSVG` — `scanDoctypeEnd` plus `getRootElementOffset`, sharing one set of rules rather than writing a second approximate scanner.
3. `feat(vite-plugin-svg-react): add an opt-in OXVG optimization pass`.

## Testing

`bun run build`, `test` (420 passing), `lint`, `tsc`, and `format:check` all clean. New coverage: default-off, default preset applied, config-object passthrough, `extend`-built config, ids preserved, class names preserved, doctype-bearing SVGs, error positions, the three strictness cases, per-file error naming, option validation, config-time config rejection, plus direct unit tests for `getRootElementOffset`.

Also verified end to end through the plugin's `load` hook on a doctype-bearing SVG: `id="brandGradient"`/`id="outline"` and `mark-fill`/`mark-dot` survive while the drawing still optimizes (`M 0,0 L 24.00000,0 L 24,24 Z` → `M0 0h24v24Z`), and the built `dist/index.js` externalizes `@oxvg/napi`.

Changeset included as a minor.

https://claude.ai/code/session_01D8Q4AktSmaUEAaAJ4pvQpM

### DIFF
--- a/.changeset/lucky-pens-repeat.md
+++ b/.changeset/lucky-pens-repeat.md
@@ -1,0 +1,39 @@
+---
+'@acusti/vite-plugin-svg-react': minor
+---
+
+Add an opt-in `optimize` option that shrinks each SVG with
+[OXVG](https://github.com/noahbald/oxvg) — the Rust, SVGO-compatible SVG
+toolchain — before it’s converted to a component. This is the replacement
+for the SVGO pass SVGR offered. It’s off by default, and `@oxvg/napi` is an
+optional peer dependency, so nothing changes unless you opt in:
+
+```
+npm install --save-dev @oxvg/napi
+```
+
+```ts
+svgReact({ optimize: true });
+```
+
+`optimize: true` is the safe setting: it drops comments, metadata and
+editor cruft, and rewrites path data and colors more compactly, but it
+leaves every `id` and `class` exactly as you authored them, so app CSS,
+`getElementById`, and `aria-labelledby` that point into an SVG keep
+working. If you’d rather have the smaller output and don’t reference those
+ids from anywhere else, pass an OXVG config object instead of `true`; the
+README shows how.
+
+Two things to know before turning it on. Optimization runs on the raw SVG
+source, ahead of everything else, so the `svg` options and the components
+you get out are unaffected. But OXVG parses the file before this plugin
+does and is stricter about XML than this plugin is: an SVG with a doctype
+that declares entities (classic Illustrator 10–CS4 did this), an
+HTML-flavored entity like `&nbsp;`, or a valueless attribute
+(`<svg hidden>`) will now fail the build. Each failure names the file and
+the line, and each file still builds with `optimize` off; the README lists
+all three.
+
+A missing `@oxvg/napi`, or a config OXVG can’t read, fails when Vite
+resolves your config rather than partway through a build, and the message
+names the option instead of whichever SVG happened to load first.
--- a/packages/vite-plugin-svg-react/README.md
+++ b/packages/vite-plugin-svg-react/README.md
@@ -65,20 +65,13 @@ below. What consumers get out of it:
   only when that round-trips, so `id="001"` stays `001` rather than turning
   into `1` and breaking the `<use href="#001">` pointing at it.
 
-One thing SVGR could do that this plugin doesn’t: run [SVGO][] optimization
-via `@svgr/plugin-svgo`. Optimizing SVGs is a build concern you can handle
-before they reach the bundler (e.g. `svgo --folder`).
-
-### Future work
-
-An optional pre-optimization pass built on [OXVG][] — the Rust,
-SVGO-compatible SVG toolchain — would fit the plugin’s all-native pipeline
-and could slot in ahead of component generation without reintroducing a JS
-compiler. If you want it, add your 👍 to [this issue][oxvg issue].
+[SVGO][]-style optimization is available as an opt-in: the `optimize`
+option below runs each SVG through [OXVG][] — the Rust, SVGO-compatible SVG
+toolchain — ahead of component generation. Its dependency is optional, so
+the default install stays dependency-free.
 
 [svgo]: https://github.com/svg/svgo
 [oxvg]: https://github.com/noahbald/oxvg
-[oxvg issue]: https://github.com/acusti/uikit/issues/422
 
 ## Usage
 
@@ -139,10 +132,10 @@ project (e.g. `src/vite-env.d.ts`):
 
 ### Options
 
-The plugin takes an optional options object with a single property: `svg`,
-which shapes the generated `<svg>` element. It supports a deliberately
-small subset of the [SVGR options][svgr options], with the same names and
-semantics (the nesting keeps the top level free for plugin-level options):
+The plugin takes an optional options object with two properties:
+`optimize`, documented below, and `svg`, which shapes the generated `<svg>`
+element. `svg` supports a deliberately small subset of the [SVGR
+options][svgr options], with the same names and semantics:
 
 ```ts
 svgReact({
@@ -160,6 +153,126 @@ svgReact({
 - `svgProps` adds extra props to the root `<svg>` (string values, or
   `{expression}` strings inserted verbatim)
 
+#### `optimize`
+
+Off by default. Set `optimize: true` to run each SVG through OXVG before
+it’s converted to a component:
+
+```
+npm install --save-dev @oxvg/napi
+```
+
+```ts
+svgReact({ optimize: true });
+```
+
+`@oxvg/napi` is an optional peer dependency: install it to use `optimize`,
+and the plugin stays dependency-free if you don’t. Optimization runs on the
+raw SVG source, so the rest of the pipeline is unchanged by it.
+
+The default preset we pass is OXVG’s default preset with `cleanupIds`
+dropped. That preset already leaves `viewBox` alone (unlike SVGO’s
+`preset-default`), and nothing it runs renames an id or a class.
+
+`cleanupIds` minifies ids and removes unreferenced ones, which is a rename
+the optimizer can’t verify: anything pointing at an id from outside the
+file — your app’s CSS, `getElementById`, an `aria-labelledby` — breaks
+silently. It’s also what makes inlined components collide, since every
+file’s ids collapse to the same `a` and `b`, and then an internal
+`<use href="#a">` or `fill="url(#a)"` resolves against whichever component
+rendered first. `@svgr/plugin-svgo` works around that by adding `prefixIds`
+on top, which renames the ids a second time and renames class names with
+them; this plugin defaults to not renaming either, so ids stay exactly as
+unique as you made them, the way inline SVG written by hand already
+behaves.
+
+If you want the smaller output and none of your ids are referenced from
+outside their file, pass OXVG’s default preset as it ships and get
+`cleanupIds` back:
+
+```ts
+import { extend } from '@oxvg/napi';
+
+svgReact({ optimize: extend({ type: 'Default' }) });
+```
+
+Passing an object hands it to OXVG’s `optimise` as-is. An OXVG config is
+the complete list of the optimizations to run, not a set of overrides on
+top of a preset, so this one runs a single optimization and nothing else:
+
+```ts
+svgReact({ optimize: { collapseGroups: { field0: true } } });
+```
+
+For the default preset with a change to it, build the config with OXVG’s
+own `extend`:
+
+```ts
+import { extend } from '@oxvg/napi';
+
+svgReact({
+    optimize: extend(
+        { type: 'Default' },
+        { removeDesc: { removeAny: true } },
+    ),
+});
+```
+
+`extend` only adds, so drop a job by leaving it out of the object you pass:
+
+```ts
+import { extend } from '@oxvg/napi';
+
+// the plugin’s own default, minus inlineStyles
+const { cleanupIds, inlineStyles, ...optimize } = extend({
+    type: 'Default',
+});
+
+svgReact({ optimize });
+```
+
+The option is typed as a plain object rather than as OXVG’s `Jobs`, so that
+this package’s types don’t reference a dependency most installs won’t have.
+For a typed config, annotate it where you write it:
+
+```ts
+import type { Jobs } from '@oxvg/napi';
+
+svgReact({ optimize: { removeDesc: { removeAny: true } } satisfies Jobs });
+```
+
+Two more things about OXVG itself:
+
+1. A job name it doesn’t recognize is ignored silently, which is what
+   `satisfies Jobs` above is for.
+2. `inlineStyles`, in the default preset, folds a rule from an SVG’s own
+   `<style>` element into a `style` attribute and drops the `class` that
+   matched it — usually a generated name like `.cls-1`, but if you style by
+   class name from your app’s CSS and the SVG carries its own rule for that
+   class, leave `inlineStyles` out as above.
+
+#### SVG files the `optimize` option rejects
+
+This plugin’s own parser is deliberately tolerant of markup that real SVG
+files contain but XML rejects. OXVG’s parser runs first and isn’t, so
+turning `optimize` on narrows what builds. Each case fails loudly, naming
+the file and the line, and each still builds with `optimize` off:
+
+- **A doctype whose internal subset declares entities the root element
+  references** — how classic Illustrator (10–CS4) wrote its Adobe
+  namespaces. OXVG rejects any document with a DTD, so the prolog is
+  blanked before it gets there, and the references outlive it:
+  `unknown entity reference 'ns_extend'`.
+- **An unknown entity name** anywhere, such as the HTML-only `&nbsp;`,
+  which XML doesn’t define. Without `optimize` these pass through
+  literally; with it, `unknown entity reference 'nbsp'`.
+- **A minimized attribute** (`<svg hidden>`), which is legal inline in HTML
+  but not in XML. Without `optimize` it reads as `hidden=""`; with it,
+  `expected '=' not '>'`.
+
+If you hit one of these, it’s the option and not your file: normalize the
+SVG once (`svgo`, or any XML formatter) or don’t use the `optimize` option.
+
 The module wrapper is fixed: a typed component that spreads its props onto
 the root `<svg>`, exported as the default export. Any other option throws
 at config time — rejecting unknown options loudly beats silently generating
@@ -174,9 +287,11 @@ replacements:
   is also the only shape `client.d.ts` types
 - `typescript`: the emitted module is compiled immediately, so this had no
   observable effect
+- `svgoConfig`: the `optimize` option above, in OXVG’s config vocabulary
+  rather than SVGO’s
 - `jsxRuntime`, `expandProps`, `titleProp`, `descProp`,
-  `replaceAttrValues`, and SVGR’s pipeline options (`plugins`, `template`,
-  `svgoConfig`): not supported
+  `replaceAttrValues`, and SVGR’s remaining pipeline options (`plugins`,
+  `template`): not supported
 
 Migrating from vite-plugin-svgr (or from this plugin’s svgr-based 0.1
 release): the old `svgrOptions` key throws with a message pointing here —
--- a/packages/vite-plugin-svg-react/src/errors.ts
+++ b/packages/vite-plugin-svg-react/src/errors.ts
@@ -1,0 +1,3 @@
+/** The message of a thrown value, which isn’t necessarily an Error. */
+export const getErrorMessage = (error: unknown) =>
+    error instanceof Error ? error.message : String(error);
--- a/packages/vite-plugin-svg-react/src/index.test.ts
+++ b/packages/vite-plugin-svg-react/src/index.test.ts
@@ -3,11 +3,31 @@ import type { ResolvedConfig } from 'vite';
 import { mkdtemp, rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+import { extend } from '@oxvg/napi';
 import { describe, expect, it } from 'vitest';
 
 import vitePluginSVGReact, { type Options } from './index.js';
 
 const SVG = '<svg xmlns="http://www.w3.org/2000/svg"><path d="M0 0h1v1H0z"/></svg>';
+
+const DTD_SVG = `\
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+    "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 4 4">
+    <path d="M 0,0 L 4,0 L 4,4 Z" fill="#ff0000"/>
+</svg>`;
+
+// classic Illustrator (10–CS4) declares its Adobe namespaces as entities in
+// the doctype’s internal subset and references them from the root element
+const ENTITY_SVG = `\
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "svg11.dtd" [
+    <!ENTITY ns_extend "http://ns.adobe.com/Extensibility/1.0/">
+]>
+<svg xmlns:x="&ns_extend;" xmlns="http://www.w3.org/2000/svg">
+    <path d="M0 0h1"/>
+</svg>`;
 
 type LoadHook = (
     this: { addWatchFile: (id: string) => void },
@@ -23,21 +43,27 @@ type ResolveIdHook = (
 
 async function loadSVGComponent({
     command,
+    fileName = 'icon.svg',
+    optimize,
+    source = SVG,
     svg,
 }: {
     command: 'build' | 'serve';
+    fileName?: string;
+    optimize?: Options['optimize'];
+    source?: string;
     svg?: Options['svg'];
 }) {
-    const plugin = vitePluginSVGReact({ svg });
+    const plugin = vitePluginSVGReact({ optimize, svg });
     const configResolved = plugin.configResolved as (
         config: Pick<ResolvedConfig, 'command'>,
-    ) => void;
-    configResolved({ command });
+    ) => Promise<void>;
+    await configResolved({ command });
 
     const directory = await mkdtemp(join(tmpdir(), 'vite-plugin-svg-react-'));
     try {
-        const filePath = join(directory, 'icon.svg');
-        await writeFile(filePath, SVG);
+        const filePath = join(directory, fileName);
+        await writeFile(filePath, source);
 
         const load = plugin.load as LoadHook;
         const watchedFiles: Array<string> = [];
@@ -229,6 +255,198 @@ describe('vite-plugin-svg-react', () => {
         expect(() =>
             vitePluginSVGReact({ svg: { svgProps: { d: 'M0 0}' } } }),
         ).not.toThrow();
+    });
+
+    it('leaves the SVG unoptimized by default', async () => {
+        const { code } = await loadSVGComponent({
+            command: 'build',
+            source: '<svg xmlns="http://www.w3.org/2000/svg"><g><path d="M 0,0 L 1,0"/></g></svg>',
+        });
+        // the group would be collapsed and the path data rewritten by OXVG
+        expect(code).toContain('"g"');
+    });
+
+    it('optimizes with OXVG’s default preset when optimize is true', async () => {
+        const { code } = await loadSVGComponent({
+            command: 'build',
+            optimize: true,
+            source: '<svg xmlns="http://www.w3.org/2000/svg"><g><path d="M 0,0 L 1,0"/></g></svg>',
+        });
+        expect(code).not.toContain('"g"');
+    });
+
+    it('runs an OXVG config object as the whole job list', async () => {
+        // an OXVG config isn’t merged into a preset — it is the job list, so
+        // a one-job config leaves everything the default preset would do
+        const { code } = await loadSVGComponent({
+            command: 'build',
+            optimize: { collapseGroups: { field0: true } },
+            source:
+                '<svg xmlns="http://www.w3.org/2000/svg">' +
+                '<desc>kept</desc><g><path d="M0 0h1"/></g></svg>',
+        });
+        expect(code).not.toContain('"g"');
+        // removeDesc is in the default preset but not in this config
+        expect(code).toContain('desc');
+    });
+
+    it('leaves ids as authored', async () => {
+        // cleanupIds is dropped from the default preset: it renames ids the
+        // optimizer can’t see all the references to (app CSS,
+        // getElementById), and minifies every file’s down to the same single
+        // letters, which is what makes inlined components collide
+        const { code } = await loadSVGComponent({
+            command: 'build',
+            optimize: true,
+            source:
+                '<svg xmlns="http://www.w3.org/2000/svg">' +
+                '<defs><linearGradient id="brandGradient"><stop offset="0"/></linearGradient></defs>' +
+                '<path id="unreferenced" d="M0 0h1v1H0z" fill="url(#brandGradient)"/></svg>',
+        });
+        expect(code).toContain('brandGradient');
+        // an unreferenced id is a hook for something outside the file, so it
+        // survives too
+        expect(code).toContain('unreferenced');
+    });
+
+    it('leaves class names alone', async () => {
+        // the classes an app’s CSS targets can’t be renamed, which rules out
+        // prefixIds (it renames class names along with ids)
+        const { code } = await loadSVGComponent({
+            command: 'build',
+            optimize: true,
+            source:
+                '<svg xmlns="http://www.w3.org/2000/svg">' +
+                '<path class="icon-fill" d="M0 0h1v1H0z"/></svg>',
+        });
+        expect(code).toContain('icon-fill');
+    });
+
+    it('optimizes SVGs that carry a doctype', async () => {
+        // OXVG refuses a document with a DTD outright, so the prolog an
+        // Illustrator export writes has to be dropped before it gets there
+        const { code } = await loadSVGComponent({
+            command: 'build',
+            optimize: true,
+            source: DTD_SVG,
+        });
+        expect(code).toContain('viewBox');
+        expect(code).not.toContain('#ff0000');
+    });
+
+    it('accepts a config built with OXVG’s extend', async () => {
+        // also a compile-time guard: `extend` returns OXVG’s `Jobs`, an
+        // interface, and TypeScript gives interfaces no implicit index
+        // signature — so typing the option as Record<string, unknown> would
+        // fail tsc right here, as it did for the README’s own examples
+        const { cleanupIds: _cleanupIds, ...optimize } = extend(
+            { type: 'Default' },
+            { removeDesc: { removeAny: true } },
+        );
+        const { code } = await loadSVGComponent({
+            command: 'build',
+            optimize,
+            source:
+                '<svg xmlns="http://www.w3.org/2000/svg">' +
+                '<desc>dropped</desc><g><path d="M0 0h1"/></g></svg>',
+        });
+        expect(code).not.toContain('desc');
+        expect(code).not.toContain('"g"');
+    });
+
+    it('fails loudly on a doctype whose internal subset declares entities', async () => {
+        // A known limitation, pinned here so it stays deliberate: the source
+        // is sliced to its root element for OXVG (which rejects any DTD),
+        // and that drops entity declarations the root element can still
+        // reference — classic Illustrator (10–CS4) declares its Adobe
+        // namespaces this way. Expanding them would be a parser of its own,
+        // to recover namespaces OXVG’s removeEditorsNSData then deletes, so
+        // this fails naming the file instead of parsing part of it.
+        await expect(
+            loadSVGComponent({
+                command: 'build',
+                optimize: true,
+                source: ENTITY_SVG,
+            }),
+        ).rejects.toThrow(/failed to optimize .*icon\.svg.*ns_extend/);
+        // without optimize the same file still builds, entity references
+        // passed through literally
+        const { code } = await loadSVGComponent({
+            command: 'build',
+            source: ENTITY_SVG,
+        });
+        expect(code).toContain('ns_extend');
+    });
+
+    it('names the SVG whose optimization failed', async () => {
+        await expect(
+            loadSVGComponent({
+                command: 'build',
+                optimize: true,
+                source: '<svg xmlns="http://www.w3.org/2000/svg"><path',
+            }),
+        ).rejects.toThrow(/failed to optimize .*icon\.svg/);
+    });
+
+    it('gives up this parser’s tolerances for non-XML markup', async () => {
+        // parse.ts accepts these deliberately (a minimized attribute is
+        // legal inline in HTML; an unknown entity name passes through
+        // literally), OXVG’s parser runs first and doesn’t. Pinned because
+        // the README promises each fails by name and builds without
+        // `optimize` — a user hitting one has to be able to tell it’s the
+        // option and not their file.
+        const cases = [
+            {
+                pattern: /expected '=' not '>'/,
+                source: '<svg xmlns="http://www.w3.org/2000/svg" hidden><path d="M0 0h1"/></svg>',
+            },
+            {
+                pattern: /unknown entity reference 'nbsp'/,
+                source: '<svg xmlns="http://www.w3.org/2000/svg"><text>a&nbsp;b</text></svg>',
+            },
+        ];
+        for (const { pattern, source } of cases) {
+            await expect(
+                loadSVGComponent({ command: 'build', optimize: true, source }),
+            ).rejects.toThrow(pattern);
+            await expect(
+                loadSVGComponent({ command: 'build', source }),
+            ).resolves.toBeTruthy();
+        }
+    });
+
+    it('reports OXVG error positions at the line the file actually has', async () => {
+        // the prolog is blanked rather than cut off, so OXVG counts from the
+        // same origin the editor does; slicing it away reported line 1 for
+        // an error five lines into the file
+        await expect(
+            loadSVGComponent({
+                command: 'build',
+                optimize: true,
+                source: ENTITY_SVG,
+            }),
+        ).rejects.toThrow(/at 5:/);
+    });
+
+    it('rejects an optimize value that is neither a boolean nor a config', () => {
+        for (const optimize of [42, 'true', ['collapseGroups'], null]) {
+            expect(() =>
+                // @ts-expect-error optimize is a boolean or an OXVG config
+                vitePluginSVGReact({ optimize }),
+            ).toThrow(/optimize must be a boolean or an OXVG config object/);
+        }
+    });
+
+    it('rejects an optimize config OXVG can’t read, at config time', async () => {
+        // OXVG’s own message names no file, and this runs before any SVG is
+        // loaded, so the error has to point back at the option itself
+        const plugin = vitePluginSVGReact({ optimize: { removeViewBox: {} } });
+        const configResolved = plugin.configResolved as (
+            config: Pick<ResolvedConfig, 'command'>,
+        ) => Promise<void>;
+        await expect(configResolved({ command: 'build' })).rejects.toThrow(
+            /OXVG rejected the optimize config/,
+        );
     });
 
     it('rejects an unmatched brace inside a regex literal or comment', () => {
--- a/packages/vite-plugin-svg-react/src/index.ts
+++ b/packages/vite-plugin-svg-react/src/index.ts
@@ -7,10 +7,22 @@ import {
     generateComponentModule,
     getComponentOptionsError,
 } from './generate.js';
+import { createOptimizer, type OptimizeConfig, type Optimizer } from './optimize.js';
 
 export type { ComponentOptions } from './generate.js';
+export type { OptimizeConfig } from './optimize.js';
 
 export type Options = {
+    /**
+     * Optimize each SVG with OXVG before it’s converted to a component.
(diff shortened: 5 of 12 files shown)


packages/vite-plugin-svg-react/src/index.ts
```diff
@@ -7,17 +7,28 @@ import {
...
+export type { OptimizeConfig } from './optimize.js';
 
 export type Options = {
+    /**
+     * Optimize each SVG with OXVG before it’s converted to a component.
+     * `true` runs OXVG’s default preset (plus prefixIds, seeded with the
+     * component name); an object is an OXVG config used as-is. Either way it
+     * needs the optional @oxvg/napi peer dependency installed.
+     * @default false
+     */
```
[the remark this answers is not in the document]
acusti: Correct, and thanks — this was a genuine miss. The default changed from "default preset + `prefixIds`" to "default preset − `cleanupIds`" during review, and `optimize.ts`, the README, the tests, and the changeset were all updated while this JSDoc was not. It's the worst place to leave stale, since it's what ships in `dist/index.d.ts` and shows up on hover in consumers' editors.

Fixed in 8a46895:

```ts
 * Optimize each SVG with OXVG before it’s converted to a component.
 * `true` runs OXVG’s default preset minus cleanupIds, so no id or class
 * name is renamed; an object is an OXVG config used as-is. Either way it
 * needs the optional @oxvg/napi peer dependency installed.
```

Verified the corrected text lands in the emitted `dist/index.d.ts`, and re-grepped the package: every remaining mention of `prefixIds` is deliberate, explaining why the default *doesn't* use it.

---

packages/vite-plugin-svg-react/src/index.test.ts
```diff
@@ -9,6 +9,15 @@ import vitePluginSVGReact, { type Options } from './index.js';
...
 const SVG = '<svg xmlns="http://www.w3.org/2000/svg"><path d="M0 0h1v1H0z"/></svg>';
 
+const DTD_SVG = [
+    '<?xml version="1.0" encoding="UTF-8"?>',
+    '<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"',
+    '    "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">',
+    '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 4 4">',
+    '    <path d="M 0,0 L 4,0 L 4,4 Z" fill="#ff0000"/>',
+    '</svg>',
+].join('\n');
```
acusti: let’s just use a template literal, e.g.;
```
const DTD_SVG = `\
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
    "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
…
```

.changeset/lucky-pens-repeat.md
```diff
@@ -0,0 +1,31 @@
...
+```ts
+svgReact({ optimize: true });
+```
+
+`optimize: true` runs OXVG’s default preset with `cleanupIds` dropped, so
+nothing it runs renames an id or a class. `cleanupIds` minifies ids and
+removes unreferenced ones, which breaks anything pointing at one from
+outside the file (app CSS, `getElementById`, an `aria-labelledby`), and
+collapses every file’s ids to the same `a` and `b`, which is what makes
+inlined components collide. `@svgr/plugin-svgo` answered that by adding
```
acusti:
```suggestion
outside the file (app CSS, `getElementById`, an `aria-labelledby`) and makes
inlined components collide. `@svgr/plugin-svgo` answered that by adding
```

.changeset/lucky-pens-repeat.md
```diff
@@ -0,0 +1,35 @@
...
+`prefixIds` on top; leaving the ids as authored keeps them as unique, or
+not, as you made them, without renaming class names too. Passing an object
+instead runs it as the whole job list, matching what OXVG’s own `optimise`
+does with a config.
```
acusti: the intended target for this changeset is someone checking on the latest features and figuring out if they want to upgrade and if/how they want to leverage the changes. i would drop the explanation of what `@svgr/plugin-svgo` does. also:

> Passing an object instead runs it as the whole job list, matching what OXVG’s own `optimise` does with a config.

isn’t very clear. i don’t know what “the whole job list” or a “job” is, so that sentence doesn’t mean anything to me.

.changeset/lucky-pens-repeat.md
```diff
@@ -0,0 +1,35 @@
...
+
+A missing `@oxvg/napi` or a config OXVG rejects fails at config time, with
+a message naming the option rather than whichever SVG loaded first. OXVG
+refuses a document with a DTD, so the plugin hands it the root element
+alone — the prolog, doctype, and header comments a modern Illustrator
+export writes are dropped first, which component generation did anyway. A
```
acusti:
```suggestion
export writes are dropped first (like what happens with component). A
```

packages/vite-plugin-svg-react/src/index.ts
```diff
@@ -7,17 +7,28 @@ import {
...
 
 export type { ComponentOptions } from './generate.js';
+export type { OptimizeConfig } from './optimize.js';
 
 export type Options = {
+    /**
+     * Optimize each SVG with OXVG before it’s converted to a component.
+     * `true` runs OXVG’s default preset minus cleanupIds, so no id or class
+     * name is renamed; an object is an OXVG config used as-is. Either way it
+     * needs the optional @oxvg/napi peer dependency installed.
```
acusti: i feel like the peer dep mention should go in a separate paragraph. not sure if it should also be separated from `* @default false` by a new line.
```suggestion
     * name is renamed; an object is an OXVG config used as-is.
     * 
     * Needs the optional @oxvg/napi peer dependency installed.
```

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,103 @@ svgReact({
...
+```
+
+```ts
+svgReact({ optimize: true });
+```
+
+`@oxvg/napi` is an optional peer dependency: install it to use `optimize`,
+and the plugin stays dependency-free if you don’t. Optimization runs on the
+raw SVG source, so the rest of the pipeline — and everything under `svg` —
+is unchanged by it.
```
acusti:
```suggestion
and the plugin stays dependency-free if you don’t. Optimization runs on the
raw SVG source, so the rest of the pipeline is unchanged by it.
```

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,103 @@ svgReact({
...
+
+`optimize: true` means OXVG’s default preset with `cleanupIds` dropped. The
+preset already leaves `viewBox` alone (unlike SVGO’s `preset-default`), and
+nothing it runs renames an id or a class.
```
acusti: i would rephrase this as something like
```suggestion
The default preset we pass is OXVG’s default preset with `cleanupIds` dropped.
That preset already leaves `viewBox` alone (unlike SVGO’s `preset-default`), and
nothing it runs renames an id or a class.
```

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,103 @@ svgReact({
...
+
+`cleanupIds` minifies ids and removes unreferenced ones, which is a rename
+the optimizer can’t verify: anything pointing at an id from outside the
+file — your app’s CSS, `getElementById`, an `aria-labelledby` — breaks
+silently. It’s also what makes inlined components collide, since every
+file’s ids collapse to the same `a` and `b`, and then an internal
+`<use href="#a">` or `fill="url(#a)"` resolves against whichever component
+rendered first. `@svgr/plugin-svgo` answered that by adding `prefixIds` on
```
acusti:
```suggestion
rendered first. `@svgr/plugin-svgo` works around that by adding `prefixIds` on
```

acusti:
Worked through the review. Per-thread replies are blocked right now — there's a pending review on the PR, and GitHub allows only one per user — so this is the one place all of it lands. Pushed as `33a2d3c` (branch restructured into three commits; details at the bottom).

Every finding reproduced before I acted on it. All eight were real.

**[Type bug: `OptimizeConfig` rejects `Jobs`](https://github.com/acusti/uikit/pull/448#discussion_r3961172188)** — confirmed with tsc against @oxvg/napi 0.0.7: `Record<string, unknown>` rejects both a `Jobs`-annotated variable and the `extend(...)` return; `object` accepts them. Took `object`. One knock-on the suggestion doesn't cover: `getDefaultConfig` destructures `cleanupIds` off `extend`'s result, which `object` isn't indexable enough for, so the internal `OXVG` type now declares `extend` as returning `Record<string, unknown>` while the exported type stays `object`. The `.d.ts` still has no `@oxvg/napi` reference. Added a test that doubles as the compile-time guard, since `bun run tsc` covers test files — it builds its config with `extend` and passes it to the plugin, so the type can't regress without failing tsc. Verified every README example compiles now, including the `jobs: Jobs` variable form that didn't before.

**[A failed load is memoized forever](https://github.com/acusti/uikit/pull/448#discussion_r3961172653)** — correct, and the comment asserted the opposite. Memoizing success only:

```ts
oxvgPromise ??= importOXVG();
try {
    return await oxvgPromise;
} catch (error) {
    oxvgPromise = null;
    throw error;
}
```

Used this shape rather than clearing inside the importing IIFE's `catch` so there's no floating `.catch()` and no ordering subtlety between the `??=` and the rejection.

**[Error positions off by the prolog's length](https://github.com/acusti/uikit/pull/448#discussion_r3961172763)** — confirmed (`1:50` for what is line 5). Adopted the blanking suggestion; OXVG accepts the leading whitespace. Pinned with a test asserting the error reports `at 5:` for a five-line-in failure.

**[`fail` typed `=> void` hides an invariant](https://github.com/acusti/uikit/pull/448#discussion_r3961172978)** — right, and the infinite-loop path is real. Typed it `=> never`; the call site satisfies it with `return fail(message)`, whose only return statement is `never`. This also matches `decodeEntities`, which already types its `fail` that way. The reasoning is now in the comment, since it's load-bearing rather than descriptive.

**[Reuse nit: `getErrorMessage`](https://github.com/acusti/uikit/pull/448#discussion_r3961173147)** — done, `src/errors.ts`, used at all three sites.

**[Lockfile drift in the feature commit](https://github.com/acusti/uikit/pull/448#discussion_r3961173348)** — confirmed: `main`'s `package.json` says 0.5.0 while `main`'s `bun.lock` says 0.4.0, so `bun install` fixed pre-existing drift. Split into its own `chore` commit.

**[Template literal for `DTD_SVG`](https://github.com/acusti/uikit/pull/448#discussion_r3961201507)** — done, and applied the same style to the new `ENTITY_SVG` fixture.

**[DTD entity declarations](https://github.com/acusti/uikit/pull/448#discussion_r3961172389)** and **[`optimize` silently tightens what's accepted](https://github.com/acusti/uikit/pull/448#discussion_r3961172557)** — both reproduced through the `load` hook, and they're the same class of problem, so I took the documenting branch for now and left both threads open, because whether to normalize instead is @acusti's call. There's a new README section, **What `optimize` stops accepting**, listing all three cases with the exact error each produces and the note that each still builds with `optimize` off; the "unchanged by it" sentence now says parsing gets stricter and points at it. Tests pin all three.

Worth knowing for that decision: OXVG's default preset deletes the Adobe namespaces the entities carry (`removeEditorsNSData`) — verified — so expanding them would recover data the optimizer then throws away. Expanding entities properly means recursion limits, external-entity refusal, and expansion caps; normalizing minimized attributes means attribute-level tokenizing. That's why I didn't do it unasked.

Branch is now `chore: sync bun.lock…` → `refactor: extract the prolog scan…` → `feat: add an opt-in OXVG optimization pass`. `bun run build`, `test` (420), `lint`, `tsc`, and `format:check` all clean.

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,103 @@ svgReact({
...
+top, but a rename to undo a rename is the wrong shape when the ids were
+fine as written — and `prefixIds` renames class names along with them.
+Leaving ids alone keeps them exactly as unique, or not, as you made them,
+which is how inline SVG written by hand already behaves.
```
acusti: we shouldn’t declare my opinion as the only opinion. instead i would say something like:

`@svgr/plugin-svgo` resolves this by doing this thing, whereas we decided to avoid the renames entirely by default. But if you want to opt-in to further optimizations that include rewriting those identifiers, you can pass an object to `optimize` instead of a boolean. So to recreate the default `@svgr/plugin-svgo` behavior, you could pass:

```ts
import { extend } from '@oxvg/napi';

<example demonstrating how to reproduce that behavior>
```

packages/vite-plugin-svg-react/src/optimize.ts
```diff
@@ -0,0 +1,111 @@
...
+    // other than the passthrough it’s documented as
+    const config = optimize === true ? getDefaultConfig(extend) : optimize;
+
+    const optimizer: Optimizer = (svg, filePath) => {
+        // OXVG refuses a document with a DTD outright, and its parser has no
+        // use for the rest of the prolog either. Component generation
+        // discards everything before the root element, so handing over just
+        // the root element loses nothing and keeps SVGs that carry the
+        // `<!DOCTYPE svg PUBLIC …>` an Illustrator export writes working.
+        const offset = getRootElementOffset(svg);
```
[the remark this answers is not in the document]
acusti: Reproduced through the `load` hook exactly as described, and taking the second branch you offered: the comment and README no longer describe the slice as lossless, and this case is named.

@acusti's call is to document rather than normalize. There's a new README section, **What `optimize` stops accepting**, covering this alongside the two sibling cases (unknown entity names, minimized attributes), each with the error it produces and the note that it still builds with `optimize` off; the "runs on the raw SVG source, so the rest of the pipeline is unchanged by it" sentence now says parsing gets stricter and points at it. A test pins the failure and pins that the same file builds without `optimize`.

Two things that informed the decision: OXVG's default preset deletes the Adobe namespaces those entities carry (`removeEditorsNSData` — verified), so expanding them recovers data the optimizer then throws away; and doing it properly means recursion limits, expansion caps, and explicit refusal of external `SYSTEM`/`PUBLIC` entities, which is a real security surface for a build plugin.

Separately, the prolog is now blanked rather than cut (per your sibling comment on error positions), so OXVG's `line:column` for this failure points at the line the file actually has.

---

packages/vite-plugin-svg-react/src/optimize.ts
```diff
@@ -0,0 +1,111 @@
...
+        try {
+            return optimise(offset === -1 ? svg : svg.slice(offset), config);
```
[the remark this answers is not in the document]
acusti: Both verified through the `load` hook: `<svg hidden>` gives `expected '=' not '>'` and `&nbsp;` gives `unknown entity reference 'nbsp'`, and both build fine with `optimize` off. Taking the documenting branch, which is @acusti's call.

The "unchanged by it" sentence now reads:

> Optimization runs on the raw SVG source, so everything under `svg` and the rest of the pipeline treat its output the same as any other file. What it does change is what counts as a valid SVG — OXVG parses before this plugin does, and it's stricter (see what `optimize` stops accepting below).

…and that new section lists these two plus the DTD-entity case, each with its exact error, closing on the point you made — that a user hitting one needs to know it's the option and not their file:

> If you hit one of these, it's the option and not your file: normalize the SVG once (`svgo`, or any XML formatter) or leave that file unoptimized.

A test pins all three: each fails by name under `optimize` and resolves without it.

On not normalizing: the entity cases would need recursion limits, expansion caps, and refusal of external entities, and the minimized-attribute case needs attribute-level tokenizing — effectively a second pass of the parser this package already has, and a second place for the two to diverge silently.

---

.changeset/lucky-pens-repeat.md
acusti: Rewrote the changeset for that reader in 8d45add. The `@svgr/plugin-svgo` explanation is gone, and so is "job list" — that sentence now says what an OXVG config *is* in plain terms rather than naming its jargon:

> `optimize: true` is the safe setting: it drops comments, metadata and editor cruft, and rewrites path data and colors more compactly, but it leaves every `id` and `class` exactly as you authored them, so app CSS, `getElementById`, and `aria-labelledby` that point into an SVG keep working. If you'd rather have the smaller output and don't reference those ids from anywhere else, pass an OXVG config object instead of `true`; the README shows how.

More generally it's reorganized around the upgrade decision — what you get, what it costs, what to know before turning it on — with the mechanism left to the README. The three cases where `optimize` makes the build stricter are stated in user terms ("an SVG with a doctype that declares entities", "a valueless attribute") rather than as parser errors.

---

.changeset/lucky-pens-repeat.md
acusti: Took the tightening ("and makes inlined components collide"), but the sentence itself is gone — your other comment asked to drop the `@svgr/plugin-svgo` explanation from the changeset and to write it for someone deciding whether to upgrade, so that whole paragraph was rewritten rather than edited. The `cleanupIds` mechanics now live only in the README.

---

.changeset/lucky-pens-repeat.md
acusti: Flagging rather than guessing: the suggestion reads "(like what happens with component)", which looks like it got cut off — I assume "component generation".

Either way that sentence is gone. Rewriting the changeset for the upgrade reader dropped the pipeline-internals framing entirely, so there's nothing left to compare component generation against. What survives of the point is stated from the user's side:

> Optimization runs on the raw SVG source, ahead of everything else, so the `svg` options and the components you get out are unaffected.

If you did want the parenthetical somewhere, the README's `optimize` section is where the prolog handling is explained now.

---

packages/vite-plugin-svg-react/src/index.ts
acusti: Applied in 8d45add, and yes to the second half — `@default false` reads better with its own break too:

```ts
    /**
     * Optimize each SVG with OXVG before it’s converted to a component.
     * `true` runs OXVG’s default preset minus cleanupIds, so no id or class
     * name is renamed; an object is an OXVG config used as-is.
     *
     * Needs the optional @oxvg/napi peer dependency installed.
     *
     * @default false
     */
```

(Dropped the trailing space after the bare `*` in your suggestion, since oxfmt strips it.)

---

packages/vite-plugin-svg-react/README.md
acusti: Applied in 8d45add — restored to your two lines.

Worth knowing what that reverts: an automated review flagged this sentence as overclaiming, because `optimize` does change what the plugin accepts (OXVG parses before `parseSVG` and is stricter), and I'd expanded it to say so and link the new **What `optimize` stops accepting** section. Your wording is defensible — the sentence is about the *pipeline*, and the pipeline genuinely is unchanged — and that section still exists a few paragraphs below with all three cases.

The tradeoff is just that the pointer to it is gone from here. Happy to add a one-line cross-reference if you want it; leaving it out for now.

---

packages/vite-plugin-svg-react/README.md
acusti: Applied verbatim in 8d45add.

---

packages/vite-plugin-svg-react/README.md
acusti: Applied in 8d45add.

---

packages/vite-plugin-svg-react/README.md
acusti: Fair — the paragraph was stating a preference as a conclusion. Rewrote it so the choice is described rather than argued, and then went looking for the example you asked for.

**It turned out `@svgr/plugin-svgo`'s behavior isn't reproducible through this option, so per your follow-up it's out of the README entirely.** Why: SVGR derived the `prefixIds` prefix from each file's name, and `optimize` resolves *one* config for every SVG, so a config-object prefix is a single literal shared by every file. That defeats the purpose — cross-file id collisions are exactly what the prefix is for. OXVG's own `prefix: { type: 'Default' }` doesn't help either; I checked, and it emits the literal string `prefix`:

```
{"type":"Default"}   -> <path id="prefix__a"/>
{"type":"Prefix",…}  -> <path id="Icon__a"/>
```

Writing that example would have shipped a config that looks like SVGR parity and quietly isn't.

So the non-dogmatic framing stayed, with an example that *is* accurate — opting back into the renames, without claiming it matches SVGR:

> That's a default, not a verdict: if you want the smaller output and none of your ids are referenced from outside their file, pass OXVG's default preset as it ships and get `cleanupIds` back.
>
> ```ts
> import { extend } from '@oxvg/napi';
>
> svgReact({ optimize: extend({ type: 'Default' }) });
> ```

Verified it runs. The `@svgr/plugin-svgo` mention in the preceding paragraph stays, with your "works around that" wording — it explains why the collision problem is known, which isn't a parity claim.

If per-file parity is something you want later, the shape that would allow it is letting `optimize` take a function of the file path. That's a real API change, so I haven't done it — flagging it as the option if it ever comes up.

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,125 @@ svgReact({
...
+`cleanupIds` minifies ids and removes unreferenced ones, which is a rename
+the optimizer can’t verify: anything pointing at an id from outside the
+file — your app’s CSS, `getElementById`, an `aria-labelledby` — breaks
+silently. It’s also what makes inlined components collide, since every
+file’s ids collapse to the same `a` and `b`, and then an internal
+`<use href="#a">` or `fill="url(#a)"` resolves against whichever component
+rendered first. `@svgr/plugin-svgo` works around that by adding `prefixIds`
+on top, which renames the ids a second time and renames class names with
+them; this plugin defaults to not renaming either, so ids stay exactly as
+unique — or not — as you made them, the way inline SVG written by hand
```
acusti: we don’t need the aside
```suggestion
unique as you made them, the way inline SVG written by hand
```

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,125 @@ svgReact({
...
+already behaves.
+
+That’s a default, not a verdict: if you want the smaller output and none of
+your ids are referenced from outside their file, pass OXVG’s default preset
+as it ships and get `cleanupIds` back.
```
acusti:
```suggestion
If you want the smaller output and none of
your ids are referenced from outside their file, pass OXVG’s default preset
as it ships and get `cleanupIds` back:
```

acusti:
Pushed `050ad4e`. Per-thread replies are blocked again by a pending review, so all three land here.

**[Drop the "— or not —" aside](https://github.com/acusti/uikit/pull/448#discussion_r3972630877)** and **[drop "That's a default, not a verdict"](https://github.com/acusti/uikit/pull/448#discussion_r3972635003)** — both applied as suggested.

**[CRLF / lone `\r` in the prolog-blanking regex](https://github.com/acusti/uikit/pull/448#discussion_r3972634184)** (Copilot) — verified, and **not making the change**, because measurement says the finding is wrong on both halves.

Fed OXVG the same prolog + a body error under each line ending, comparing the current `/[^\n]/g` against the proposed `/[^\n\r]/g`. The `<svg>` is on file line 5 in all three:

| line endings | current `/[^\n]/g` | proposed `/[^\n\r]/g` |
| --- | --- | --- |
| LF | `at 5:50` ✅ | `at 5:50` |
| CRLF | `at 5:50` ✅ | `at 5:50` |
| lone CR | `at 1:153` ❌ | `at 1:153` ❌ |

CRLF is already correct: the `\r` sits at the *end* of a line, so blanking it moves nothing — the `\n` still terminates the line and the following line's columns are untouched. And for lone `\r`, preserving it changes nothing, because the position never depended on the blanking in the first place.

Confirmed that separately with a file that has **no prolog at all**, so the plugin does zero blanking and OXVG sees the bytes as they are — entity on file line 2:

```
LF    -> at 2:12
CRLF  -> at 2:12
CR    -> at 1:52
```

So OXVG counts lines by `\n` alone. A lone-`\r` file reports line 1 whether or not this code touches it; the regex can't cause it and can't fix it.

Worth adding: `parseSVG`'s own `fail()` counts the same way (`if (input[position] === '\n')`), so on such a file the optimize path and the non-optimize path agree, rather than one being a regression against the other. Lone-`\r` line endings are classic Mac OS, which ended in 2001.

Rather than a no-op change that would imply the problem was fixed, the reasoning is now in the comment above the regex, so this doesn't get re-raised:

```ts
// Everything but a line feed, so blanking the prolog preserves line and
// column. \n only is deliberate, and enough: under CRLF the \r sits at the
// end of a line, so blanking it moves nothing, and OXVG counts lines by \n
// alone — a file using lone \r line endings reports line 1 whether or not
// the prolog is touched. parseSVG’s own positions count the same way, so the
// two agree on such a file rather than one being a regression on the other.
```

420 tests, lint, tsc, format:check clean.

---

packages/vite-plugin-svg-react/src/errors.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * The message of a thrown value, which isn’t necessarily an Error: a plugin
+ * wrapping a failure from elsewhere (oxc, OXVG’s native binding) has no
+ * guarantee about what was thrown, and `String(error)` is what reads best
+ * for the rest.
```
acusti: this is a very familiar pattern, we don’t need this much explanation
```suggestion
 * The message of a thrown value, which isn’t necessarily an Error
```

packages/vite-plugin-svg-react/src/index.ts
```diff
@@ -7,17 +7,30 @@ import {
...
+     */
+    optimize?: boolean | OptimizeConfig;
     /**
      * Options shaping the generated <svg> element (dimensions, icon, svgProps)
      * with the same semantics as svgr’s options of the same names.
      */
     svg?: ComponentOptions;
 };
 
+const OPTION_NAMES: ReadonlySet<string> = new Set(['optimize', 'svg']);
```
acusti, line 33: let’s move this to directly below `virtualModuleFilter`

packages/vite-plugin-svg-react/src/optimize.ts
```diff
@@ -0,0 +1,152 @@
...
+// minifies ids and drops unreferenced ones. That’s a rename the optimizer
+// can’t verify — anything pointing at an id from outside the file (app CSS,
+// getElementById, an aria-labelledby) breaks silently — and collapsing every
+// file’s ids to the same `a` and `b` makes duplicate DOM ids likely as soon
+// as two components are inlined on one page. @svgr/plugin-svgo answered that
+// second problem by adding prefixIds on top, but a rename to undo a rename
+// is the wrong shape when the ids were fine as authored, and prefixIds
+// renames class names along with them. Leaving ids alone keeps them exactly
+// as unique — or not — as the author made them, which is what inline SVG
+// written by hand already does.
```
acusti: too much for a code comment, plus this is already captured in multiple other places
```suggestion
// What `optimize: true` runs: OXVG’s default preset minus cleanupIds, which
// minifies ids and drops unreferenced ones. That’s a rename the optimizer
// can’t verify — anything pointing at an id from outside the file (app CSS,
// getElementById, an aria-labelledby) breaks silently — and collapsing every
// file’s ids to the same `a` and `b` makes duplicate DOM ids likely as soon
// as two components are inlined on one page.
```

packages/vite-plugin-svg-react/src/optimize.ts
```diff
@@ -0,0 +1,152 @@
...
+ * Build the optimizer for an `optimize` option that isn’t `false`, loading
+ * @oxvg/napi and running the resolved config once so that a missing package
+ * or a config OXVG rejects fails here — where the message can name the
+ * option — rather than on the first `.svg?react` import, where it would name
+ * whichever SVG happened to get there first.
+ */
+export async function createOptimizer(
+    optimize: OptimizeConfig | true,
+): Promise<Optimizer> {
+    const { extend, optimise } = await loadOXVG();
```
acusti, line 98: these two variables with an extremely subtle spelling difference (`optimize` + `optimise`) are too easy to confuse. let’s rename `optimize` to `config` or `options` or similar.

packages/vite-plugin-svg-react/README.md
```diff
@@ -139,10 +133,11 @@ project (e.g. `src/vite-env.d.ts`):
...
 
-The plugin takes an optional options object with a single property: `svg`,
-which shapes the generated `<svg>` element. It supports a deliberately
-small subset of the [SVGR options][svgr options], with the same names and
-semantics (the nesting keeps the top level free for plugin-level options):
+The plugin takes an optional options object with two properties:
+`optimize`, documented below, and `svg`, which shapes the generated `<svg>`
+element. `svg` supports a deliberately small subset of the [SVGR
+options][svgr options], with the same names and semantics (the nesting
+keeps the top level free for plugin-level options like `optimize`):
```
acusti: time to drop this aside, it’s already been demonstrated earlier in the same sentence
```suggestion
options][svgr options], with the same names and semantics:
```

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,125 @@ svgReact({
...
+the complete list of the optimizations to run, not a set of overrides on
+top of a preset, so this one runs a single optimization and nothing else:
+
+```ts
+svgReact({ optimize: { collapseGroups: { field0: true } } });
+```
+
+For the default preset with a change to it, build the config with OXVG’s
+own `extend` — which is also how you get `cleanupIds` (or `prefixIds`)
+back:
```
acusti: we just covered how to “get `cleanupIds` back” two paragraphs above
```suggestion
For the default preset with a change to it, build the config with OXVG’s
own `extend`:
```

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,125 @@ svgReact({
...
+svgReact({ optimize: { removeDesc: { removeAny: true } } satisfies Jobs });
+```
+
+Two more things about OXVG itself. A job name it doesn’t recognize is
+ignored silently, which is what `satisfies Jobs` above is for. And
+`inlineStyles`, in the default preset, folds a rule from an SVG’s own
+`<style>` element into a `style` attribute and drops the `class` that
+matched it — usually a generated name like `.cls-1`, but if you style by
+class name from your app’s CSS and the SVG carries its own rule for that
+class, leave `inlineStyles` out as above.
```
acusti: let’s make this a real list
```suggestion
Two more things about OXVG itself:

1. A job name it doesn’t recognize is
ignored silently, which is what `satisfies Jobs` above is for
2. `inlineStyles`, in the default preset, folds a rule from an SVG’s own
`<style>` element into a `style` attribute and drops the `class` that
matched it — usually a generated name like `.cls-1`, but if you style by
class name from your app’s CSS and the SVG carries its own rule for that
class, leave `inlineStyles` out as above
```

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,125 @@ svgReact({
...
+
+#### What `optimize` stops accepting
```
acusti: i would find this clearer if it was more along the lines of
```suggestion
#### Incompatibilities introduced by using the `optimize` option
```

i’m not at all attached to that wording, but “stops accepting” made me think this was gonna be about config options

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +155,125 @@ svgReact({
...
+  `unknown entity reference 'ns_extend'`.
+- **An unknown entity name** anywhere, such as the HTML-only `&nbsp;`,
+  which XML doesn’t define. Without `optimize` these pass through
+  literally; with it, `unknown entity reference 'nbsp'`.
+- **A minimized attribute** (`<svg hidden>`), which is legal inline in HTML
+  but not in XML. Without `optimize` it reads as `hidden=""`; with it,
+  `expected '=' not '>'`.
+
+If you hit one of these, it’s the option and not your file: normalize the
+SVG once (`svgo`, or any XML formatter) or leave that file unoptimized.
```
acusti: how would the user “leave that file unoptimized”? if there’s a way we should cover that above. if not, we should change “or leave that file unoptimized” to “or don’t use the `optimize` option”.

packages/vite-plugin-svg-react/src/optimize.ts
acusti: Agreed — `optimize` vs `optimise` one line apart is a trap. Renamed the parameter to `options` in 9c637c7 (`config` was already taken by the derived value on the next line):

```ts
export async function createOptimizer(
    options: OptimizeConfig | true,
): Promise<Optimizer> {
    const { extend, optimise } = await loadOXVG();
    const config = options === true ? getDefaultConfig(extend) : options;
```

---

packages/vite-plugin-svg-react/README.md
acusti: There isn't a way — `optimize` is a single plugin-level option with no include/exclude, so it applies to every `.svg?react` import. Took your second reading, and said the missing capability out loud so nobody goes looking for it:

> If you hit one of these, it's the option and not your file: normalize the SVG once (`svgo`, or any XML formatter) or don't use the `optimize` option. It applies to every `.svg?react` import — there's no way to exempt a single file.

---

packages/vite-plugin-svg-react/README.md
acusti: Good catch — "stops accepting" does read like it's about options. Went with:

> #### SVG files the `optimize` option rejects

Same intent as yours, shorter, and "SVG files" makes it unambiguous up front. Nothing linked to the old anchor, so the rename is self-contained. Happy to take your exact wording if you prefer it.

---

packages/vite-plugin-svg-react/src/optimize.ts
acusti: Applied your suggestion verbatim in 9c637c7 — the `@svgr/plugin-svgo` / prefixIds half is gone. It was saying the same thing as the README section on `cleanupIds` and the changeset, so the code comment now stops at why the job is dropped.

---

packages/vite-plugin-svg-react/src/index.ts
acusti: Moved in 9c637c7. It needed a one-line comment above it to sit there: `perfectionist/sort-modules` sorts alphabetically within comment-delimited partitions, so `OPTION_NAMES` after `VIRTUAL_PREFIX`/`virtualModuleFilter` fails the rule unless it starts its own partition.

```ts
const virtualModuleFilter = /^\0vite-plugin-svg-react:/;
// every option this plugin takes, so a typo at the top level is rejected
const OPTION_NAMES: ReadonlySet<string> = new Set(['optimize', 'svg']);
```

---

packages/vite-plugin-svg-react/src/errors.ts
acusti: Applied in 9c637c7 — collapsed to the single line (kept it as `/** … */` on one line, with the trailing period).

---

packages/vite-plugin-svg-react/README.md
acusti: Applied in 9c637c7.

---

packages/vite-plugin-svg-react/README.md
acusti: Applied in 9c637c7.

---

packages/vite-plugin-svg-react/README.md
acusti: Applied in 9c637c7, as a numbered list. Indented the continuation lines so each item stays one paragraph under oxfmt, and kept the trailing periods.

---

packages/vite-plugin-svg-react/src/index.ts
acusti: Reverted in 6a42da7 — `OPTION_NAMES` is back at the top of the const block, with no comment added. The lint rule made the move cost more than it was worth.

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -65,20 +65,14 @@ below. What consumers get out of it:
...
-via `@svgr/plugin-svgo`. Optimizing SVGs is a build concern you can handle
-before they reach the bundler (e.g. `svgo --folder`).
-
-### Future work
-
-An optional pre-optimization pass built on [OXVG][] — the Rust,
-SVGO-compatible SVG toolchain — would fit the plugin’s all-native pipeline
-and could slot in ahead of component generation without reintroducing a JS
-compiler. If you want it, add your 👍 to [this issue][oxvg issue].
+What SVGR did that this plugin doesn’t do by default: run [SVGO][]
```
acusti: SVGR ran SVGO by default?

packages/vite-plugin-svg-react/README.md
```diff
@@ -160,6 +154,128 @@ svgReact({
...
+  which XML doesn’t define. Without `optimize` these pass through
+  literally; with it, `unknown entity reference 'nbsp'`.
+- **A minimized attribute** (`<svg hidden>`), which is legal inline in HTML
+  but not in XML. Without `optimize` it reads as `hidden=""`; with it,
+  `expected '=' not '>'`.
+
+If you hit one of these, it’s the option and not your file: normalize the
+SVG once (`svgo`, or any XML formatter) or don’t use the `optimize` option.
+It applies to every `.svg?react` import — there’s no way to exempt a single
+file.
```
acusti: no need to call this out
```suggestion
```

packages/vite-plugin-svg-react/src/parse.ts
```diff
@@ -35,6 +35,90 @@ const isXMLCharacter = (codePoint: number) =>
...
+// declarations with their own '>'s), and skipping comment and
+// processing-instruction bodies so their contents can’t derail the
+// quote/bracket tracking. Returns -1 when the declaration is unterminated,
+// reporting it through `fail` first when one is given.
+//
+// `fail` returns never, and that’s load-bearing rather than descriptive:
+// parseSVG advances past the returned offset unconditionally, so a callback
+// that returned instead of throwing would send it back to offset 0 and loop
+// over the document forever. Typing it this way makes the compiler hold the
+// caller to it.
```
acusti, line 81: this is cool, i've `never` seen this technique before

packages/vite-plugin-svg-react/README.md
acusti: Good question — the sentence was overstated, and I've fixed it in 4970c96. Checked the actual packages rather than going from memory:

| package | `defaultPlugins` | runs SVGO? |
| --- | --- | --- |
| `@svgr/cli` 8.1.0 | `[svgo, jsx, prettier]` | **yes** |
| `@svgr/webpack` 8.1.0 | `[svgo, jsx]` | **yes** |
| `vite-plugin-svgr` 5.2.0 | `[jsx]` | no |
| `@acusti/vite-plugin-svg-react` **0.1.0** | imports `@svgr/plugin-jsx` only | no |

`@svgr/core` on its own runs no plugins at all — its `svgo: true` default only gates the plugin *if* the caller puts it in the list, which the Vite ones don't.

So: true of SVGR's CLI and webpack loader, false of the Vite path, and false of this plugin's own 0.1 — which made the old wording read like `optimize` was restoring something readers had lost. It now says:

> SVGR's CLI and webpack loader ran SVGO optimization by default, via `@svgr/plugin-svgo`. The Vite path never did: vite-plugin-svgr loads only `@svgr/plugin-jsx`, as did this plugin's own svgr-based 0.1. The `optimize` option below is the equivalent, running OXVG …

---

packages/vite-plugin-svg-react/README.md
acusti: Removed in 4970c96.

---

packages/vite-plugin-svg-react/src/parse.ts
acusti: The trick is that `return fail(message)` is what makes the arrow's inferred return type `never` — `fail` returns `never`, and a function whose only `return` yields `never` is itself `never`. Dropping the `return` and just calling `fail(message)` infers `void`, which isn't assignable to the parameter type, so the compiler catches the one mistake that matters.

Worth noting `decodeEntities` in the same file already typed its `fail` parameter this way — I'd broken the pattern by loosening it to `=> void` when extracting the scanner, which is exactly what the review caught.

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -65,20 +65,16 @@ below. What consumers get out of it:
...
-SVGO-compatible SVG toolchain — would fit the plugin’s all-native pipeline
-and could slot in ahead of component generation without reintroducing a JS
-compiler. If you want it, add your 👍 to [this issue][oxvg issue].
+SVGR’s CLI and webpack loader ran [SVGO][] optimization by default, via
+`@svgr/plugin-svgo`. The Vite path never did: [vite-plugin-svgr][] loads
+only `@svgr/plugin-jsx`, as did this plugin’s own svgr-based 0.1. The
+`optimize` option below is the equivalent, running [OXVG][] — the Rust,
+SVGO-compatible SVG toolchain — ahead of component generation. It’s opt-in
+and its dependency is optional, so the default install stays
+dependency-free.
```
acusti: no need for this history either, just what's relevant to the library consumer, namely that the plugin supports opt-in SVGO-style optimization via the optional dependency on OXVG

acusti: Cut to that in ccc7fce — the whole SVGR/0.1 comparison is gone:

> [SVGO][]-style optimization is available as an opt-in: the `optimize` option below runs each SVG through [OXVG][] — the Rust, SVGO-compatible SVG toolchain — ahead of component generation. Its dependency is optional, so the default install stays dependency-free.

Which also retires the question you asked two comments up: with the "what SVGR did by default" framing gone, there's no claim left to get wrong. Both link references are still used, so the definitions below still resolve.

---

acusti (commented):
LGTM

PR #450, opened 2026-09-11
feat(vite-plugin-svg-react): per-file id prefixes and optimize include/exclude

acusti:
## Summary

Two changes to the `optimize` option added in #448, each its own commit and changeset (both minor).

### 1. `optimize: true` runs the full default preset, with ids prefixed per file

0.3.0 dropped `cleanupIds` from the default because it minifies every file's ids down to the same `a` and `b`, so components inlined together collide. Measured on a real icon set, that gave up about a tenth of the savings.

OXVG's `prefixIds` is the fix, but `optimise` takes no path, so its `{ type: 'Default' }` prefix degrades to the literal string `prefix` in every file. The `load` hook does have the path, so the optimizer now resolves the config per file: the default preset plus `prefixIds` with a prefix of the file's sanitized base name and a 4-char sha256 of its path relative to the vite root, joined to the id by `_`:

```html
<!-- src/icons/arrow.svg -->
<linearGradient id="arrow-58f3_a">…<path fill="url(#arrow-58f3_a)"/>
```

- Hashing the root-relative path keeps output identical across machines; same-named files in different directories get different prefixes.
- Class names stay unprefixed. In-file references, `aria-labelledby` included, are rewritten.
- A job list's own `{ type: 'Default' }` prefix gets the same treatment; explicit and `None` prefixes pass through.
- The cost is what 0.3.0 avoided: an id referenced only from outside its file is unreferenced to `cleanupIds` and removed. The README says so, notes that the minified suffix isn't stable across edits, and gives the drop-`cleanupIds` recipe.

Verified in Chromium that `<use href>`, `url(#…)` fills, clip paths and masks all resolve through the prefixed ids. OXVG runs jobs in its own fixed order (`cleanupIds` before `prefixIds`) regardless of key order.

### 2. Object form of `optimize` is now `{ exclude, include, jobs }`

For the files that shouldn't go through `cleanupIds` (an illustration whose ids app CSS animates) while the icon set next to them should.

```ts
svgReact({ optimize: { exclude: ['src/illustrations/**'] } });
```

- `include`/`exclude` are Vite `FilterPattern`s, resolved against the vite root with `createFilter`, so semantics match every other plugin's. Excluded SVGs still become components from their source as written.
- `jobs` is the OXVG job list the option itself used to be. Every key is optional; no `jobs` means the default preset.
- **Breaking for 0.3.0 configs:** a job list passed as `optimize` directly now throws with a message naming `jobs`, rather than sniffing the object for OXVG keys.

## Test plan

- [x] New tests: per-file prefix and `url(#…)` rewrite, same-named files in different dirs differ, Default prefix in a job list resolved per file, explicit prefix left alone, `getIdPrefix` stability and sanitizing, include-only, exclude-only, legacy job-list shape throws, non-object `jobs` throws
- [x] `bun run --filter '@acusti/vite-plugin-svg-react' test` (93 passing), `tsc`, `build`
- [x] `bun run lint`, `bun run format:check`

https://claude.ai/code/session_01Lj5sTV4JRWs9DP2uXfjKEd

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lj5sTV4JRWs9DP2uXfjKEd)_

<a href="https://cubic.dev/pr/acusti/uikit/pull/450?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/optimize-include-exclude.md
+++ b/.changeset/optimize-include-exclude.md
@@ -1,0 +1,20 @@
+---
+'@acusti/vite-plugin-svg-react': minor
+---
+
+The object form of `optimize` is now `{ exclude, include, jobs }`, each
+optional. `include` and `exclude` narrow which SVGs the pass runs on — a
+glob, a RegExp, or an array of either, matched against the path relative to
+the vite root with Vite’s `createFilter` semantics — and `jobs` is the OXVG
+job list that the 0.3 release took as the option’s value itself. With no
+`jobs`, the default preset runs, so
+`optimize: { exclude: ['src/illustrations/**'] }` is the full default minus
+those files, which still become components from their source as written.
+The `true` short form is not affected by this reshaping (what it runs
+changed separately; see the entry above). A job list with `cleanupIds` gets
+the per-file `prefixIds` unless it brings its own, so a customized preset
+stays as collision-safe as the default, and dropping `cleanupIds` keeps ids
+as authored.
+
+Breaking: passing an OXVG job list as `optimize` directly now throws with a
+message pointing at `jobs`; move it there.
--- a/.changeset/prefix-ids-per-file.md
+++ b/.changeset/prefix-ids-per-file.md
@@ -15,7 +15,7 @@ The cost is that an id referenced only from outside its file (app CSS,
 `cleanupIds` can tell, and is removed. To keep ids as authored, pass the
 default preset minus `cleanupIds` as a config object; the README shows how.
 
-For config objects, a `prefixIds` prefix of `{ type: 'Default' }` is now
-resolved to the same per-file prefix rather than reaching OXVG as the
-literal `prefix`; an explicit prefix or `{ type: 'None' }` is passed
+For a job list of your own, a `prefixIds` prefix of `{ type: 'Default' }`
+is now resolved to the same per-file prefix rather than reaching OXVG as
+the literal `prefix`; an explicit prefix or `{ type: 'None' }` is passed
 through unchanged.
--- a/.changeset/prefix-ids-per-file.md
+++ b/.changeset/prefix-ids-per-file.md
@@ -1,0 +1,21 @@
+---
+'@acusti/vite-plugin-svg-react': minor
+---
+
+`optimize: true` now runs OXVG’s full default preset, `cleanupIds`
+included, and prefixes each file’s ids with a prefix derived from the file:
+its base name, a `-`, and a 4-character hash of its path relative to the
+vite root, with `_` between the prefix and the id (`arrow-3f2a_a`). Ids
+come out minified, and unique across components inlined on one page, where
+0.3.0 left them untouched to avoid the collisions `cleanupIds` alone
+causes. Class names still aren’t renamed.
+
+The cost is that an id referenced only from outside its file (app CSS,
+`getElementById`, an `aria-labelledby` elsewhere) is unreferenced as far as
+`cleanupIds` can tell, and is removed. To keep ids as authored, pass the
+default preset minus `cleanupIds` as a config object; the README shows how.
+
+For config objects, a `prefixIds` prefix of `{ type: 'Default' }` is now
+resolved to the same per-file prefix rather than reaching OXVG as the
+literal `prefix`; an explicit prefix or `{ type: 'None' }` is passed
+through unchanged.
--- a/packages/vite-plugin-svg-react/README.md
+++ b/packages/vite-plugin-svg-react/README.md
@@ -172,20 +172,11 @@ svgReact({ optimize: true });
 and the plugin stays dependency-free if you don’t. Optimization runs on the
 raw SVG source, so the rest of the pipeline is unchanged by it.
 
-`optimize: true` runs OXVG’s default preset, which already leaves `viewBox`
-alone (unlike SVGO’s `preset-default`), plus one job of our own on top of
-it. The preset’s `cleanupIds` minifies ids and removes unreferenced ones,
-and it minifies every file’s ids down to the same `a` and `b` — so as soon
-as two components are inlined on one page, an internal `<use href="#a">` or
-`fill="url(#a)"` resolves against whichever component rendered first.
-SVGO’s `prefixIds` exists for exactly this, prefixing each file’s ids with
-its file name, but OXVG’s `optimise` takes no path, so its `Default` prefix
-is the literal string `prefix` in every file and the ids collide the same.
-This plugin does have the path, so it runs `prefixIds` with a prefix
-derived from each file: the file’s base name plus a 4-character hash of its
-path relative to the vite root, separated from the id by `_`. Two files
-named `arrow.svg` in different directories get different prefixes, and the
-output is the same on every machine:
+`optimize: true` runs OXVG’s default preset, which leaves `viewBox` alone
+(unlike SVGO’s `preset-default`), plus `prefixIds`. Ids come out minified
+and prefixed with the file’s base name and a 4-character hash of its path
+relative to the vite root, so they stay unique when components are inlined
+together and come out the same on every machine:
 
 ```
 // src/icons/arrow.svg, as authored
@@ -197,73 +188,118 @@ output is the same on every machine:
 <path fill="url(#arrow-58f3_a)"/>
 ```
 
-Class names aren’t prefixed (`prefixClassNames: false`), and nothing in the
-default preset renames a class, so the classes your app’s CSS targets stay
-as you wrote them. Ids don’t: an id referenced only from outside the file —
-your app’s CSS, `getElementById`, an `aria-labelledby` on another element —
-is unreferenced as far as `cleanupIds` can tell, and removed. References
-inside the file, `aria-labelledby` included, are rewritten with the id. The
-prefix is stable for a given path, but the minified part isn’t: which id
-becomes `a` depends on the order of references in the file, so an edit to
-the SVG can reassign it. If you reference ids from outside their files,
-don’t bind to the optimized names; drop `cleanupIds` from the preset and
-leave `prefixIds` out, which keeps every id exactly as unique as you made
-it, the way inline SVG written by hand already behaves:
+Every reference inside the file — `href="#…"`, `url(#…)`, `aria-labelledby`
+— is rewritten to match. Class names are left as you wrote them, with one
+exception: a class the SVG’s own `<style>` element styles is folded into a
+`style` attribute and dropped (`inlineStyles`, see the end of this
+section).
+
+To avoid issues, don’t reference an id inside an SVG from outside it (app
+CSS, `getElementById`, an `aria-labelledby` on another element). An `id`
+that isn’t referenced anywhere else in the file is removed, and the
+minified part isn’t stable — which id becomes `a` depends on the order of
+references in the file, so an edit to the SVG can reassign it. For an id
+the rest of your app needs, put it on the component instead. Props spread
+onto the root `<svg>`, so an `id` prop lands there, and everything inside
+is reachable from it by class:
+
+```tsx
+import Hero from './hero.svg?react';
+
+function Banner() {
+    return <Hero id="hero-art" />;
+}
+```
+
+```css
+#hero-art .wheel {
+    animation: spin 4s linear infinite;
+}
+```
+
+If an SVG’s ids have to stay as authored, leave `cleanupIds` out and pass
+the rest of the preset as `jobs`:
 
 ```ts
 import { extend } from '@oxvg/napi';
 
-// the default preset minus cleanupIds, and no prefixIds
-const { cleanupIds, ...optimize } = extend({ type: 'Default' });
+// the default preset minus cleanupIds
+const { cleanupIds, ...jobs } = extend({ type: 'Default' });
 
-svgReact({ optimize });
+svgReact({ optimize: { jobs } });
 ```
 
-Passing an object hands it to OXVG’s `optimise` as-is. An OXVG config is
-the complete list of the optimizations to run, not a set of overrides on
-top of a preset, so this one runs a single optimization and nothing else:
+Or keep the default and leave those files out of the pass. An object form
+of the option takes `include` and `exclude` patterns — a glob, a RegExp, or
+an array of either, matched with Vite’s `createFilter` against each SVG’s
+path relative to the vite root (`icons/star.svg`), for RegExps and globs
+alike — alongside the `jobs` to run. Each is optional: with no `jobs`, the
+default preset runs; with no `include`, every SVG that `exclude` doesn’t
+match is optimized. Excluded SVGs still become components, just from their
+source as written:
 
 ```ts
-svgReact({ optimize: { collapseGroups: { field0: true } } });
+svgReact({
+    optimize: {
+        exclude: ['src/illustrations/**', /\.animated\.svg$/],
+        include: 'src/**',
+    },
+});
 ```
 
-The one value the plugin fills in is a `prefixIds` prefix of
-`{ type: 'Default' }`, which is meaningless to `optimise` without a path:
-it resolves to the same per-file prefix `optimize: true` uses, so this is
-the default preset with `prefixIds` also prefixing class names:
+`jobs` is handed to OXVG’s `optimise` as-is. An OXVG job list is the
+complete list of the optimizations to run, not a set of overrides on top of
+a preset, so this one runs a single optimization and nothing else:
+
+```ts
+svgReact({ optimize: { jobs: { collapseGroups: { field0: true } } } });
+```
+
+The plugin owns one key in the list, `prefixIds`. A job list with
+`cleanupIds` gets the per-file `prefixIds` that `optimize: true` uses
+unless it brings its own, so a customized preset stays as collision-safe as
+the default one, and a list without `cleanupIds` isn’t prefixed. In a
+`prefixIds` of your own, a `prefix` of `{ type: 'Default' }` — which is
+meaningless to `optimise` without a path — resolves to that same per-file
+prefix, so this is the default preset with `prefixIds` also prefixing class
+names:
 
 ```ts
 import { extend } from '@oxvg/napi';
 
 svgReact({
-    optimize: extend(
-        { type: 'Default' },
-        {
-            prefixIds: {
-                delim: '_',
-                prefix: { type: 'Default' },
-                prefixClassNames: true,
-                prefixIds: true,
+    optimize: {
+        jobs: extend(
+            { type: 'Default' },
+            {
+                prefixIds: {
+                    delim: '_',
+                    prefix: { type: 'Default' },
+                    prefixClassNames: true,
+                    prefixIds: true,
+                },
             },
-        },
-    ),
+        ),
+    },
 });
 ```
 
 A `prefix` of `{ type: 'Prefix', field0: 'app' }` or `{ type: 'None' }` is
 left alone.
 
-For the default preset with a change to it, build the config with OXVG’s
+For the default preset with a change to it, build the job list with OXVG’s
 own `extend`:
 
 ```ts
 import { extend } from '@oxvg/napi';
 
 svgReact({
-    optimize: extend(
-        { type: 'Default' },
-        { removeDesc: { removeAny: true } },
-    ),
+    optimize: {
+        jobs: extend(
+            { type: 'Default' },
+            { removeDesc: { removeAny: true } },
+        ),
+    },
 });
 ```
 
@@ -272,21 +308,22 @@ svgReact({
 ```ts
 import { extend } from '@oxvg/napi';
 
-// the default preset minus inlineStyles (and, since this is a config
-// object, without the prefixIds that `optimize: true` adds)
-const { inlineStyles, ...optimize } = extend({ type: 'Default' });
+// the default preset minus inlineStyles
+const { inlineStyles, ...jobs } = extend({ type: 'Default' });
 
-svgReact({ optimize });
+svgReact({ optimize: { jobs } });
 ```
 
-The option is typed as a plain object rather than as OXVG’s `Jobs`, so that
+`jobs` is typed as a plain object rather than as OXVG’s `Jobs`, so that
 this package’s types don’t reference a dependency most installs won’t have.
-For a typed config, annotate it where you write it:
+For a typed job list, annotate it where you write it:
 
 ```ts
 import type { Jobs } from '@oxvg/napi';
 
-svgReact({ optimize: { removeDesc: { removeAny: true } } satisfies Jobs });
+svgReact({
+    optimize: { jobs: { removeDesc: { removeAny: true } } satisfies Jobs },
+});
 ```
 
 Two more things about OXVG itself:
--- a/packages/vite-plugin-svg-react/README.md
+++ b/packages/vite-plugin-svg-react/README.md
@@ -172,30 +172,51 @@ svgReact({ optimize: true });
 and the plugin stays dependency-free if you don’t. Optimization runs on the
 raw SVG source, so the rest of the pipeline is unchanged by it.
 
-The default preset we pass is OXVG’s default preset with `cleanupIds`
-dropped. That preset already leaves `viewBox` alone (unlike SVGO’s
-`preset-default`), and nothing it runs renames an id or a class.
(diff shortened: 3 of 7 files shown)


.changeset/prefix-ids-per-file.md
```diff
@@ -0,0 +1,25 @@
...
+root, separated by `_` (`arrow-3f2a_a`). `cleanupIds` was dropped from the
+default in 0.3.0 because it minifies every file’s ids down to the same `a`
+and `b`, so components inlined together collide; measured on a real icon
+set that was giving up about a tenth of the savings. OXVG’s `prefixIds`
+can’t fix it alone — `optimise` takes no path, so its default prefix is the
+literal string `prefix` in every file — but the plugin has the path, so it
+resolves the prefix per file. Class names still aren’t renamed.
+
+The cost is the one 0.3.0 avoided: an id referenced only from outside its
+file (app CSS, `getElementById`, an `aria-labelledby` elsewhere) is
```
acusti:
```suggestion
The cost is that an id referenced only from outside its
file (app CSS, `getElementById`, an `aria-labelledby` elsewhere) is
```

packages/vite-plugin-svg-react/src/index.ts
```diff
@@ -31,6 +55,14 @@ export type Options = {
...
 
 const OPTION_NAMES: ReadonlySet<string> = new Set(['optimize', 'svg']);
+const OPTIMIZE_OPTION_NAMES: ReadonlySet<string> = new Set([
+    'exclude',
+    'include',
+    'jobs',
+]);
+
+const isPlainObject = (value: unknown): value is Record<string, unknown> =>
+    typeof value === 'object' && value !== null && !Array.isArray(value);
```
acusti: maybe this should be a `function isPlainObject` at the bottom of the file and just let it be hoisted so it’s available everywhere. doesn’t seem to me to fit here next to these constants

packages/vite-plugin-svg-react/src/index.ts
```diff
@@ -82,22 +114,47 @@ export default function vitePluginSVGReact(options: Options = {}): Plugin {
...
-    ) {
+    if (typeof optimize !== 'boolean' && !isPlainObject(optimize)) {
         throw new Error(
-            'vite-plugin-svg-react: optimize must be a boolean or an OXVG config object ' +
-                '(see the README’s Options section).',
+            'vite-plugin-svg-react: optimize must be a boolean or an object with ' +
+                'exclude, include, and/or jobs (see the README’s Options section).',
         );
     }
+    if (isPlainObject(optimize)) {
```
acusti, line 126: let’s separate this block with an empty line
```suggestion

    if (isPlainObject(optimize)) {
```

acusti:
Review round addressed and folded into the two feature commits (b7c5eee, 6b22770). Per-thread replies are blocked by a pending review draft on this PR, so the dispositions are here instead:

- **Changeset wording (acusti's suggestion):** applied.
- **RegExp `include`/`exclude` never matching (cubic):** confirmed. With `resolve: root`, `createFilter` roots string globs but tests a RegExp against the absolute id. The filter now runs on the root-relative posix path with `resolve: false`, so globs and RegExps see the same path, matching what the docs claimed. The include test now covers a RegExp pattern.
- **README migration note (cubic):** the "Upgrading from 0.3" paragraph now covers both the `jobs` shape and the `optimize: true` behavior change, pointing at the drop-`cleanupIds` recipe and at `exclude`.
- **Test recomputing the prefix with `getIdPrefix` (cubic):** the plugin test now pins the literal `brand-bc08_a`, and the `getIdPrefix` unit test pins a literal hash, so a change to the hash source or length fails.
- **Measurement in the changeset (cubic):** dropped; it stays in the PR description.

---

acusti:
Second round, folded into 3fb0de4 (per-thread replies still blocked by the pending review draft):

- **`isPlainObject` placement (acusti):** now a hoisted `function` at the bottom of the file.
- **Blank line before the object-form validation block (acusti):** applied.
- **Error message names `optimize.jobs` for `optimize: true` (Copilot):** it now names `optimize` for the boolean and `optimize.jobs` for a job list.
- **README and JSDoc say root-relative but the code matched the absolute path (Copilot, two threads):** the code was the wrong side and 6b22770 already fixed it, filtering the root-relative posix path with `resolve: false`. The README paragraph and JSDoc now show the path shape they match (`icons/star.svg`) and drop the "same semantics as every other plugin" claim, since other plugins test RegExps against absolute ids.

---

packages/vite-plugin-svg-react/src/optimize.ts
```diff
@@ -41,15 +52,72 @@ type OXVG = {
...
+    typeof value === 'object' &&
+    value !== null &&
+    typeof (value as { prefix?: unknown }).prefix === 'object' &&
+    (value as { prefix: null | object }).prefix !== null;
+
+// The config’s prefixIds job if it asks for the default prefix, which is the
+// one this plugin resolves per file. `optimise` takes no path, so OXVG’s
+// `{ type: 'Default' }` — “derive it from the file name”, in SVGO’s
+// definition — degrades there to the literal string `prefix`, and every
+// file’s ids collide identically. The `load` hook does have the path.
```
acusti: no need for all the background, that’s captured already in custom messages and PR descriptions 
```suggestion
// the config’s prefixIds job if it asks for the default prefix, which is the
// one this plugin resolves per file
```

packages/vite-plugin-svg-react/README.md
```diff
@@ -172,51 +172,120 @@ svgReact({ optimize: true });
...
-rendered first. `@svgr/plugin-svgo` works around that by adding `prefixIds`
-on top, which renames the ids a second time and renames class names with
-them; this plugin defaults to not renaming either, so ids stay exactly as
-unique as you made them, the way inline SVG written by hand already
-behaves.
-
-If you want the smaller output and none of your ids are referenced from
-outside their file, pass OXVG’s default preset as it ships and get
-`cleanupIds` back:
+`optimize: true` runs OXVG’s default preset, which already leaves `viewBox`
```
acusti: pre-existing, but let’s drop `already`
```suggestion
`optimize: true` runs OXVG’s default preset, which leaves `viewBox`
```

packages/vite-plugin-svg-react/README.md
```diff
@@ -172,51 +172,120 @@ svgReact({ optimize: true });
...
+as two components are inlined on one page, an internal `<use href="#a">` or
+`fill="url(#a)"` resolves against whichever component rendered first.
+SVGO’s `prefixIds` exists for exactly this, prefixing each file’s ids with
+its file name, but OXVG’s `optimise` takes no path, so its `Default` prefix
+is the literal string `prefix` in every file and the ids collide the same.
+This plugin does have the path, so it runs `prefixIds` with a prefix
+derived from each file: the file’s base name plus a 4-character hash of its
+path relative to the vite root, separated from the id by `_`. Two files
+named `arrow.svg` in different directories get different prefixes, and the
+output is the same on every machine:
```
acusti: this is way too much about the problems that led to this solution. a reader of this doc just wants to know “What does this do and how do I use this and what gotchas do I have to worry about?”

this should just document what the plugin does and what the user can expect to see and what they should look out for.

packages/vite-plugin-svg-react/README.md
```diff
@@ -172,51 +172,120 @@ svgReact({ optimize: true });
...
+as you wrote them. Ids don’t: an id referenced only from outside the file —
+your app’s CSS, `getElementById`, an `aria-labelledby` on another element —
+is unreferenced as far as `cleanupIds` can tell, and removed. References
+inside the file, `aria-labelledby` included, are rewritten with the id. The
+prefix is stable for a given path, but the minified part isn’t: which id
+becomes `a` depends on the order of references in the file, so an edit to
+the SVG can reassign it. If you reference ids from outside their files,
+don’t bind to the optimized names; drop `cleanupIds` from the preset and
+leave `prefixIds` out, which keeps every id exactly as unique as you made
+it, the way inline SVG written by hand already behaves:
```
acusti: let’s say “Ids are renamed:…” (instead of “Ids don’t:…”)

also, i was thinking the most common practical way to handle ids that should be external reference-able is to define them as top-level props on the SVG React component, e.g.
```tsx
import MyIcon from '~/icon.svg';

function MyComponent() {
    return <MyIcon id="my-icon-id" />;
}
```
or maybe with a `useId()`:

```tsx
import { useId } from 'react';
import SomeText from '~/some-text.svg';

function MyComponent() {
    const myId = useId();
    return (
        <>
            <SomeText id={myId} />;
            <button aria-labelledby={myId} />;
        </>
    );
}
```
i know that last example doesn’t make sense, but hopefully the concept is clear.

packages/vite-plugin-svg-react/README.md
```diff
@@ -225,24 +294,32 @@ svgReact({
...
+});
 ```
 
+Upgrading from 0.3: a job list passed as `optimize` itself throws with a
+message pointing at `jobs`, and `optimize: true` itself changed — it ran
+the preset without `cleanupIds` and renamed nothing, and now minifies and
+prefixes ids as described above. If your app references ids inside an SVG,
+pass the `jobs` recipe above (the preset minus `cleanupIds`) or `exclude`
+those files.
+
```
acusti: we don’t need to include an “Upgrading from 0.3” note
```suggestion
```

packages/vite-plugin-svg-react/src/optimize.ts
acusti: Applied, in 80b4195.

---

packages/vite-plugin-svg-react/README.md
acusti: Applied, in 0c4ecc8.

---

packages/vite-plugin-svg-react/README.md
acusti: Rewrote the section in 0c4ecc8. It now goes: what `optimize: true` runs and what ids look like afterwards (with the before/after example), what gets rewritten and what doesn't, then the gotcha ("Ids are renamed: don't reference them from outside") with the id-prop pattern as the fix, and the keep-ids-as-authored recipe last. The SVGO/OXVG history is gone.

---

packages/vite-plugin-svg-react/README.md
acusti: Done in 0c4ecc8. "Ids are renamed:" leads the paragraph, and the id-prop pattern is now the documented fix: `<Hero id="hero-art" />` lands on the root `<svg>` via the props spread, and a CSS example targets `#hero-art .wheel`, which works because classes aren't renamed. One sentence points at `useId()` for the multi-instance case. The upgrading note lists it ahead of the `jobs` recipe and `exclude`.

---

packages/vite-plugin-svg-react/README.md
acusti: Removed, in 31fa10c.

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -172,51 +172,130 @@ svgReact({ optimize: true });
...
+Every reference inside the file — `href="#…"`, `url(#…)`, `aria-labelledby`
+— is rewritten to match. Class names are left as you wrote them.
+
+Ids are renamed: don’t reference an id inside an SVG from outside it (app
+CSS, `getElementById`, an `aria-labelledby` on another element). An id
+nothing in the file references is removed, and the minified part isn’t
+stable — which id becomes `a` depends on the order of references in the
+file, so an edit to the SVG can reassign it. For an id the rest of your app
+needs, put it on the component instead. Props spread onto the root `<svg>`,
+so an `id` prop lands there, and everything inside is reachable from it by
```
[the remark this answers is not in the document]
acusti: Fair. In 5fc4052 the "class names are left as you wrote them" sentence now carries the `inlineStyles` exception and points at the note at the end of the section, and the id-prop paragraph references it.

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -172,51 +172,133 @@ svgReact({ optimize: true });
...
+section).
+
+Ids are renamed: don’t reference an id inside an SVG from outside it (app
+CSS, `getElementById`, an `aria-labelledby` on another element). An id
+nothing in the file references is removed, and the minified part isn’t
+stable — which id becomes `a` depends on the order of references in the
+file, so an edit to the SVG can reassign it. For an id the rest of your app
+needs, put it on the component instead. Props spread onto the root `<svg>`,
+so an `id` prop lands there, and everything inside is reachable from it by
+class (the `inlineStyles` exception above aside):
```
acusti: re-reading this, we just covered that ids are renamed (“Ids come out minified and prefixed” and “Every reference inside the file — `href="#…"`, `url(#…)`, `aria-labelledby` — is rewritten to match.”), so the “Ids are renamed” is unnecessary. instead, we should say something like:

```suggestion
To avoid issues, don’t reference an id inside an SVG from outside it (app
CSS, `getElementById`, an `aria-labelledby` on another element). An `id`
that isn’t referenced anywhere else in the file is removed, and the minified part isn’t
stable — which id becomes `a` depends on the order of references in the
file, so an edit to the SVG can reassign it. For an id the rest of your app
needs, put it on the component instead. Props spread onto the root `<svg>`,
so an `id` prop lands there, and everything inside is reachable from it by
class:
```

acusti: Applied, in 3462dd8 (rewrapped to the column width).

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -172,51 +172,133 @@ svgReact({ optimize: true });
...
+```
+
+```css
+#hero-art .wheel {
+    animation: spin 4s linear infinite;
+}
+```
+
+Use `useId()` for the value when the component renders more than once on a
+page.
```
acusti: actually i think we should drop this, it’s self-evident

acusti: Dropped, in c3758cd.

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -172,51 +172,130 @@ svgReact({ optimize: true });
...
 
 ```ts
-svgReact({ optimize: { collapseGroups: { field0: true } } });
+svgReact({ optimize: { jobs: { collapseGroups: { field0: true } } } });
+```
+
+The one value the plugin fills in is a `prefixIds` prefix of
+`{ type: 'Default' }`, which is meaningless to `optimise` without a path:
+it resolves to the same per-file prefix `optimize: true` uses, so this is
+the default preset with `prefixIds` also prefixing class names:
```
acusti: double-colon

acusti: Split into two sentences, in 6b796d9.

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -225,22 +304,23 @@ svgReact({
 ```ts
 import { extend } from '@oxvg/napi';
 
-// the plugin’s own default, minus inlineStyles
-const { cleanupIds, inlineStyles, ...optimize } = extend({
-    type: 'Default',
-});
+// the default preset minus inlineStyles (and, since this is a job list of
+// your own, without the prefixIds that `optimize: true` adds)
```
acusti: oh i didn’t realize this. customizing `jobs` necessarily means disabling our cleanupIds/prefixIds functionality? that’s not great.

acusti: Not quite "disabling", but close enough to be a footgun: a `jobs` object is passed verbatim, so `extend({ type: 'Default' })` keeps `cleanupIds` and has no `prefixIds`, and ids collide again. The prefix is recoverable today by adding `prefixIds: { prefix: { type: 'Default' }, … }`, which the plugin resolves per file, but that's a thing to remember on every custom list.

Two ways to close it:

1. **Let `jobs` also be a function of the plugin's default list:** `jobs: (defaults) => { const { inlineStyles, ...jobs } = defaults; return jobs; }`, where `defaults` is the preset plus the per-file `prefixIds`. Customizing keeps the prefixing unless you drop it on purpose; an object stays verbatim for full control. Explicit, no sniffing, and the "minus one job" recipes in the README become one-liners.
2. **Inject `prefixIds` into any `jobs` object that lacks a `prefixIds` key.** Simpler for users, but it makes the passthrough lie, and opting out means spelling a full `prefixIds` job with `prefixIds: false`.

I'd take 1. Say the word and I'll push it; not pushing an API change on my own.

---

acusti: hmmmm… i’m leaning 2, but i can see why explicit is safer. but it seems like 1 would require renaming the option. `jobs` as a function is weird.

acusti: Agreed that `jobs` as a function is weird. A tighter version of 2 that I think fixes its opt-out problem:

**When a `jobs` list has `cleanupIds` and no `prefixIds` key, the plugin adds its per-file `prefixIds` alongside it.** Anything else is verbatim. An explicit `prefixIds` is respected as now (with a `Default` prefix resolved per file).

The tie to `cleanupIds` is the point: prefixing exists to make minified ids unique, so it rides along with the job that minifies them and only that one. That gives:

- `extend({ type: 'Default' }, { removeDesc: … })` → still prefixed, no extra work.
- `const { inlineStyles, ...jobs } = extend(…)` → still prefixed.
- `const { cleanupIds, ...jobs } = extend(…)` → ids as authored, nothing else to know. No `prefixIds: false` spelling needed.
- Prefix without minify: write `prefixIds` with a `Default` prefix yourself.

One rule in the README: "a job list with `cleanupIds` gets the per-file `prefixIds` unless it brings its own". Want me to push that?

---

acusti: yeah that sounds right

acusti: Pushed in 130f770. A job list with `cleanupIds` and no `prefixIds` key gets the plugin's per-file `prefixIds`; everything else is verbatim. New test covers `extend({ type: 'Default' }, { removeDesc: … })` coming out prefixed, and the existing minus-`cleanupIds` test still shows ids as authored. README recipes lost their "and no prefixIds" caveats, the `jobs` paragraph states the rule, and the changeset mentions it.

---

.changeset/optimize-include-exclude.md
```diff
@@ -0,0 +1,19 @@
...
+glob, a RegExp, or an array of either, matched against the path relative to
+the vite root with Vite’s `createFilter` semantics — and `jobs` is the OXVG
+job list that the 0.3 release took as the option’s value itself. With no
+`jobs`, the default preset runs, so
+`optimize: { exclude: ['src/illustrations/**'] }` is the full default minus
+those files, which still become components from their source as written.
+`optimize: true` is unchanged. A job list with `cleanupIds` gets the
+per-file `prefixIds` unless it brings its own, so a customized preset stays
+as collision-safe as the default, and dropping `cleanupIds` is all it takes
+to keep ids as authored.
```
acusti:
```suggestion
as collision-safe as the default, and dropping `cleanupIds` keeps ids as authored.
```

acusti: Applied, in 70f491b (rewrapped to the column width).

---

packages/vite-plugin-svg-react/src/index.ts
```diff
@@ -173,3 +240,7 @@ export default function vitePluginSVGReact(options: Options = {}): Plugin {
         },
     };
 }
+
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+    return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
```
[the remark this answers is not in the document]
acusti: Confirmed and fixed in adf425a: the guard now accepts only objects whose prototype is `Object.prototype` or null, and the rejection test includes a `Date`. Checked first that OXVG's `extend` returns plain objects, so `jobs` built with it still passes.

---

.changeset/optimize-include-exclude.md
```diff
@@ -0,0 +1,19 @@
...
+
+The object form of `optimize` is now `{ exclude, include, jobs }`, each
+optional. `include` and `exclude` narrow which SVGs the pass runs on — a
+glob, a RegExp, or an array of either, matched against the path relative to
+the vite root with Vite’s `createFilter` semantics — and `jobs` is the OXVG
+job list that the 0.3 release took as the option’s value itself. With no
+`jobs`, the default preset runs, so
+`optimize: { exclude: ['src/illustrations/**'] }` is the full default minus
+those files, which still become components from their source as written.
+`optimize: true` is unchanged. A job list with `cleanupIds` gets the
```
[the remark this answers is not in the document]
acusti: Fair, the sentence was ambiguous since both changesets land in one release. Reworded in fa5c1c5: the `true` short form is not affected by this reshaping, and what it runs changed separately, per the sibling entry.

---

.changeset/prefix-ids-per-file.md
```diff
@@ -0,0 +1,23 @@
+---
+'@acusti/vite-plugin-svg-react': minor
+---
+
+`optimize: true` now runs OXVG’s full default preset, `cleanupIds`
+included, and prefixes each file’s ids with a prefix derived from the file:
+its base name plus a 4-character hash of its path relative to the vite
+root, separated by `_` (`arrow-3f2a_a`). `cleanupIds` was dropped from the
+default in 0.3.0 because it minifies every file’s ids down to the same `a`
+and `b`, so components inlined together collide. OXVG’s `prefixIds` can’t
```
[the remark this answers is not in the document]
acusti: Trimmed in f7ab2ee: the OXVG internals are gone, and the paragraph now states the user-visible effect (minified, per-file unique ids; classes untouched; what 0.3.0 did instead).

---

packages/vite-plugin-svg-react/README.md
```diff
@@ -172,51 +172,134 @@ svgReact({ optimize: true });
...
-Passing an object hands it to OXVG’s `optimise` as-is. An OXVG config is
-the complete list of the optimizations to run, not a set of overrides on
-top of a preset, so this one runs a single optimization and nothing else:
+`jobs` is handed to OXVG’s `optimise` as-is. An OXVG job list is the
+complete list of the optimizations to run, not a set of overrides on top of
+a preset, so this one runs a single optimization and nothing else:
 
 ```ts
-svgReact({ optimize: { collapseGroups: { field0: true } } });
+svgReact({ optimize: { jobs: { collapseGroups: { field0: true } } } });
```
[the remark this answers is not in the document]
acusti: Not adding this: the README had an "Upgrading from 0.3" note and the author removed it deliberately. The migration is documented in the changeset, and the config-time error names `jobs` directly.

---

.changeset/prefix-ids-per-file.md
```diff
@@ -0,0 +1,21 @@
+---
+'@acusti/vite-plugin-svg-react': minor
+---
+
+`optimize: true` now runs OXVG’s full default preset, `cleanupIds`
+included, and prefixes each file’s ids with a prefix derived from the file:
+its base name plus a 4-character hash of its path relative to the vite
+root, separated by `_` (`arrow-3f2a_a`). Ids come out minified, and unique
```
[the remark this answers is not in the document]
acusti: Correct, fixed in 06463b1: the sentence now says `-` joins the base name to the hash and `_` sits between the prefix and the id.

---

packages/vite-plugin-svg-react/src/index.ts
```diff
@@ -173,3 +240,11 @@ export default function vitePluginSVGReact(options: Options = {}): Plugin {
         },
     };
 }
+
+// a plain object only: a Date or a class instance has no own keys to fail
+// the unknown-keys check, so it would silently pass as `{}`
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+    if (typeof value !== 'object' || value === null) return false;
```
acusti: i would find this most readable if we started with a nullish check, then the `typeof`
```suggestion
    if (value == null || typeof value !== 'object') return false;
```

acusti: Applied, in 79a0cdc.

---

acusti (commented):
LGTM

